PR #9, opened 2017-07-17
Sparse sheets fix

SereneAnt:
Changes:
- fixed column names generation and schema inference for sparse sheets (with lots of nulls);
`row.cellIterator()` did not work good in this case.
- library versions updated.
- two more optional options (hehe) added for selective load (startColumn/endCoumn, inclusive)


nightscape:
Hi @SereneAnt,

big :+1: for the PR and even adding tests!
Merging this now and releasing a new version.

nightscape:
0.8.4 should appear on Sonatype within a few hours.

Issue #10, opened 2017-08-10
option for inserting userSchema is not possible


raviuiit:
HI I am tryin to read a date field from excel, but your API doesn't provide support for same. So i try to provide user provided schema as shown in the example in readme, but still not able to pass as schema is an object not a string. 
val df = sqlContext.read
    .format("com.crealytics.spark.excel")
    .option("location", "Worktime.xlsx")
    .option("sheetName", "Daily")
    .option("useHeader", "true")
    .option("treatEmptyValuesAsNulls", "true")
    **# .option("userSchema", StructType(???)) // Optional// Not possible to pass schema as string**
    .option("inferSchema", "true")
    .option("addColorColumns", "true")
    .option("startColumn", 0) // Optional
    .option("endColumn", 99) // Optional
    .load()

nightscape:
Can you provide more detail what is happening? Do you get the corresponding column as a String or is an exception thrown or something completely different?

raviuiit:
while reading a date field from excel we normally get a numeric value from POI API. So for fetching the same we need to use 
`if ((HSSFDateUtil.isCellDateFormatted(cell)))  `
and then format the cell using date formatter.

SO when i am using Your API i am getting the date field as Numeric as expeccted. So now when i tried looking into the code i found that if i pass an schema object then i would be able to get the field as date.
Schema:
`StructType schema = DataTypes.createStructType(new StructField[] {DataTypes.createStructField("Created", DataTypes.DateType, true) });`

But in your API i am not able to pass schema as shown in the example posted in readme.md on github.

So please provide an option to read date field.

nightscape:
Hi @raviuiit,

could you locally build https://github.com/crealytics/spark-excel/pull/13 and see if that works for you?

DraftBeer99:
I had the same problem, when i read a date cell to spark,  What kind of StructType do I use,i get a double value,Is there any way to deal with this?

nightscape:
I just released 0.8.5 which should fix this. I had to drop Scala 2.10 support because a dependency does not provide Scala 2.10 builds.
I hope everybody managed to switch to 2.11 already :wink:

raviuiit:
I am trying to do this for a field 'Created' in Excel
`StructType schema = DataTypes.createStructType(new StructField[] {DataTypes.createStructField("Created", DataTypes.DateType, true) });`

[...]

PR #11, opened 2017-08-12
Feature/excelwithschema

ergunbaris:
To be able to provide external schema I found this approach to be the smoothest. In our case we would like to definitely keep track of input and output data schemas and using avro for this purpose. After these changes I have tested successfully an integration with the latest snapshot of spark-excel library on my fork which accepted external schema with .format("com.crealytics.spark.excel.schema") I hope this helps others as well. FYI I have seen 1 user also asking for the same support on your repo.


ergunbaris:
FYI:
https://github.com/crealytics/spark-excel/issues/10

nightscape:
Hi @ergunbaris,

thanks for the PR! I created a version with a smaller diff here:
https://github.com/crealytics/spark-excel/pull/13
Could you build that and see if it works for you?

ergunbaris:
The feature is added by crealytics

PR #12, opened 2017-08-12
Feature/excelwithschema2

ergunbaris:
This is the other alternative solution to extracting data from an excel with a StructType schema provided by the user. This one has only one DefaultSource which is the default one and if no schema is provided it will default to existing logic.


ergunbaris:
This one breaks the current inferschema if no schema is provided! So withdrawing  it!

PR #13, opened 2017-08-14
Actually allow overriding the schema

nightscape:


ergunbaris:
I had this solution as an alternative which than I've withdrawn today because I have discovered that if I didn't provide a schema the existing routine was broken with an Analysis Exception that schema was not provided. I see that you have provided SchemaRelationProvider as a mixin trait where as I did extend from it instead of BaseRelation and thus always expected an schema. 

I assume you probably have done your local dev tests with no schema and everything is fine.

I will test with schema and let you know soon. Thx.

ergunbaris:
Sadly this solution fails. Need to read Spark documentation more about this but here is the current situation 
1-) if you provide a DefaultSource that extends RelationProvider;  **any** schema providing with schema method -> fails.
2-) if you provide a DefaultSource that extends SchemaRelationProvider **no** schema case in other words not using schema method -> fails. 
with AnalysisException
This was the reason I did provide 2 different namespaces and 2 different DefaultSources. Will try to provide you with unit tests also if I can find time.

nightscape:
Hmm, that is strange. I tried to replicate the [behavior of spark-csv](https://github.com/databricks/spark-csv/blob/master/src/main/scala/com/databricks/spark/csv/DefaultSource.scala#L28-L54) which is also described e.g. [here](https://mapr.com/blog/spark-data-source-api-extending-our-spark-sql-query-engine/).
The [pattern matching code that actually calls the `createRelation` methods](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/DataSource.scala#L332-L344) imo should actually work with that style of implementation:
```scala
// if a schema is provided then this case should match because our DefaultSource is a SchemaRelationProvider
      case (dataSource: SchemaRelationProvider, Some(schema)) =>
        dataSource.createRelation(sparkSession.sqlContext, caseInsensitiveOptions, schema)
// if no schema is provided then this case should match because our DefaultSource is also a RelationProvider
      case (dataSource: RelationProvider, None) =>
        dataSource.createRelation(sparkSession.sqlContext, caseInsensitiveOptions)
// this should not happen because the previous cases should have matched
      case (_: SchemaRelationProvider, None) =>
        throw new AnalysisException(s"A schema needs to be specified when using $className.")
// this should not happen because the previous cases should have matched
      case (dataSource: RelationProvider, Some(schema)) =>
        val baseRelation =
          dataSource.createRelation(sparkSession.sqlContext, caseInsensitiveOptions)
        if (baseRelation.schema != schema) {
          throw new AnalysisException(s"$className does not allow user-specified schemas.")
        }
        baseRelation
```
Maybe I missed something, but I'm quite sure that it shouldn't be necessary to create two DefaultSources.

ergunbaris:
Quick Correction while configuring my environment in development I am using pom versions to point to latest version. Yesterday night I realized that I have a bigger version for your library on my local trials so I have corrected to point to the latest of yours . Good news TEST PASS. 

By looking at how Csv is implemented I can understand your approach. Mine was the simplest quickest solution which provided me a way to tackle around adding schema support.

So if everything allright on your side I can verify I can get it to working for user provided schema.

AakashBasu:
Hi,

I need an urgent help in this, can you please help me?

Here is the problem statement clearly defined -
 https://stackoverflow.com/questions/45713699/inferschema-using-spark-read-formatcom-crealytics-spark-excel-is-inferring-d

While reading the Excel data, it is a forced inferSchema, which I do not want as this JAR is converting any numeric value to double, including the date, there's no way I am finding it possible to revert it back after that.

Tried with building the schema dynamically and passing it over to the df while reading the data and removing the inferSchema option, which is erroring out by saying -

**py4j.protocol.Py4JJavaError: An error occurred while calling o72.load.
: java.lang.IllegalArgumentException: Parameter "inferSchema" is missing in options.**

Thanks,
Aakash.

AakashBasu:
@nightscape ,

I did the same, downloaded the 0.9.0 and tried to not use that option at all, but I am getting this error as I'm having to use 

df = spark.read.format("com.crealytics.spark.excel").\
...     option("location", "D:\\Users\\aabasu\\Desktop\\GameStop\\Spark_Programming\\TmpData\\SPRING202020REPORT.xlsm").\
...     option("sheetName", "202020Details").\
...     option("treatEmptyValuesAsNulls", "true").\
...     option("addColorColumns", "false").\
...    option("spark.read.simpleMode","true"). \
... option("useHeader", "true").\
...     load("com.databricks.spark.csv")

I get this Error - "py4j.protocol.Py4JJavaError: An error occurred while calling o48.load.
: java.io.FileNotFoundException: File com.databricks.spark.csv does not exist"


And if I do not use the load part with this, that DF isn't of any use -

Eg: df = spark.read.format("com.crealytics.spark.excel").\
...     option("location", "D:\\Users\\aabasu\\Desktop\\GameStop\\Spark_Programming\\TmpData\\SPRING202020REPORT.xlsm").\
...     option("sheetName", "202020Details").\
...     option("treatEmptyValuesAsNulls", "true").\
...     option("addColorColumns", "false").\
...    option("spark.read.simpleMode","true"). \
... option("useHeader", "true")
>>> df.show()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'DataFrameReader' object has no attribute 'show'

So, I think this is something which you need to see into in the latest release of 0.9.0, what I did for a solution is I reverted back to the 0.8.6 edition and used  **option("inferSchema", "false").\** and used a dynamic Schema and it worked. :)

nightscape:
You're using it wrong, just check the README again or read what the error message actually says :wink:
You have to `.load("D:\Users\aabasu\Desktop\Spark_Programming\TmpData\ABC.xlsm")` and not `.load("com.databricks.spark.csv")`.

AakashBasu:
Hey,

I think I'm not doing it incorrectly. I think you're not getting the context I'm trying to explain. See,

1) While using your package, we give **.option("location", "D:\Users\aabasu\Desktop\Spark_Programming\TmpData\ABC.xlsm").** instead of **.load("D:\Users\aabasu\Desktop\Spark_Programming\TmpData\ABC.xlsm")** to read the source dataset.

2) What I was trying to mean is, along with the mandatory options, if we do not give this **.load("com.databricks.spark.csv")**, what happens is, while hitting any action, it throws an error. The error is always like, 

> Traceback (most recent call last):
> File "", line 1, in 
> AttributeError: 'DataFrameReader' object has no attribute 'show'

3) But in the older version, i.e., 0.8.6, it worked fine with the load of csv along with **.format("com.crealytics.spark.excel")**. But in the latest version 0.9.0, the above issue in point 2 is coming up.

I somehow managed with the 0.8.6 itself and couldn't use 0.9.0 due to this issue.

Hope this clarifies the exact problem statement to you.

Thanks,
Aakash.

nightscape:
I think there's still some misunderstanding:
The `.option("location", filePath)` version does not exist anymore in 0.9.0,
 see the [CHANGELOG](https://github.com/crealytics/spark-excel/blob/master/CHANGELOG.md#086) and the [updated README](https://github.com/crealytics/spark-excel#create-a-dataframe-from-an-excel-file). The `.load(filePath)` method is there instead to specify the path to the file you want to read, not the package you want to use (and `.load("com.databricks.spark.csv")` is even the wrong package name).
If you don't call `.load()` or `.load(filePath)` at the end of the method chain you will get a `DataFrameReader` instead of a `DataFrame`.
For the package you need to use `.format("com.crealytics.spark.excel")` as you mention in 3.

Please copy and paste the code from the current README, adapt it to your needs and if **that** still doesn't work create a new issue.

PR #14, opened 2017-08-15
Feature added to be able support external schema as StructType

ergunbaris:
This is the same pull request as feature/excelwithschema branch only that I got rid of the crazy commit logs which were repeating them selves because I had to amend some commit comments and rebase it. Now everything changed for this feature is gathered under 1 commit.


ergunbaris:
The feature is added by crealytics

Issue #18, opened 2017-09-03
ExcelFileSaver.toCell() doesn't handle null values in DataFrame


slehan:
When a `DataFrame` row has a null value in one of the columns, a `scala.MatchError` exception gets thrown by `toCell`. I am writing with the following `write` calls:

```
    dataFrame.write
      .format("com.crealytics.spark.excel")
      .option("sheetName", config.sheetName)
      .option("treatEmptyValuesAsNulls", true)
      .option("useHeader", true)
      .mode("overwrite")
      .save(config.path)
```

This is reproducible by loading in an Excel or CSV file with blanks in any of the columns and attempting to save it. I added the `treatEmptyValuesAsNulls` just to make sure I wasn't missing an option but looking at the code it seems that a case needs to be added to handle if `a == null`. It is also possible I'm doing something wrong here, and if so, I apologize!

nightscape:
No, you're not doing anything wrong and your diagnose is correct: The `toCell` method does not handle `null` yet https://github.com/crealytics/spark-excel/blob/master/src/main/scala/com/crealytics/spark/excel/ExcelFileSaver.scala#L28-L40
Do you dare creating a PR for this?

slehan:
I'll take a look at it today and see if I can't fix it!

slehan:
@nightscape can we just assume that if it's null, we can just write an empty string to the Excel spreadsheet?

nightscape:
You might want to use [`Cell.Empty`](https://github.com/norbert-radyk/spoiwo/blob/master/src/main/scala/com/norbitltd/spoiwo/model/Cell.scala#L28) in that case.
Currently that's the same as an empty string, but we'd be on the safe side if they decide to change that at one point.

nightscape:
Closed by https://github.com/crealytics/spark-excel/pull/20
Version 0.9.1 should be on Maven central within a few hours.

PR #20, opened 2017-09-04
Null values fix

slehan:
Let me know what you think about the test I made, it does pass all other tests, including the one I made.

### DIFF
--- a/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
+++ b/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
@@ -88,8 +88,15 @@ class IntegrationSuite extends FunSuite with PropertyChecks with DataFrameSuiteB
   test("handles null values correctly") {
     forAll(rowsGen, MinSuccessful(20)) { rows =>
       val expected = spark.createDataset(rows).toDF
+
+      // We need two dataframes, one with null values, one with empty strings, this is because we want ExcelFileSaver to
+      // write an empty string, if there's a null in that column. expectedWithEmptyStr is what the dataframe should look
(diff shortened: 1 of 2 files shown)


slehan:
Submitted to fix issue #18

src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
```diff
@@ -83,4 +84,40 @@ class IntegrationSuite extends FunSuite with PropertyChecks with DataFrameSuiteB
...
+      val expectedWithNull = expected.withColumn("aString", lit(null: String))
+      // Generate the same DataFrame but with empty strings
+      val expectedWithEmptyStr = expected.withColumn("aString", lit("": String))
+      // Set the schema so that aString is nullable
+      expectedWithEmptyStr.schema.fields.update(6, new StructField("aString", DataTypes.StringType, true))
+
+      val tempFile: File = File.createTempFile("spark_excel_null_test_", ".xlsx")
+      val fileName = tempFile.getAbsolutePath
+
+      expectedWithNull.write
```
nightscape: Could you extract a method like `writeThenRead(df: DataFrame): DataFrame` from line 104-118 and also reuse that in the existing test-case?
Just to keep the duplication down and make the actual intent of the test clearer.

ManiRz: When we convert a excel file into a dataframe, if the first row contains null in the last column, it is excluding the entire column. Is there any way to fix this? Thanks.

nightscape: Can you create a new issue for this?

PR #23, opened 2017-09-07
allow timestamp format to be specified if desired

davidw76:
Before this change, timestamps would only be parsed if they were in the default format `yyyy-[m]m-[d]d hh:mm:ss[.f...]`

Otherwise, an exception was thrown.

With this change, an optional "timestampFormat" parameter can be given with a different format string, e.g. `hh:mm:ss:SSS`

If no timestampFormat is given, then the behavior remains the same as before


davidw76:
spotted something wrong..

PR #24, opened 2017-09-07
allow timestamp format to be specified

davidw76:
Before this change, timestamps would only be parsed if they were in the default format `yyyy-[m]m-[d]d hh:mm:ss[.f...]`. Otherwise, an exception was thrown.

With this change, an optional _timestampFormat_ parameter can be given with a different format string, for example `hh:mm:ss:SSS`

If no _timestampFormat_ is given, then the behavior remains the same as before

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ df.write
   .format("com.crealytics.spark.excel")
   .option("sheetName", "Daily")
   .option("useHeader", "true")
+  .option("timestampFormat", "HH:mm:ss") // Optional, default: yyyy-MM-dd HH:mm:ss.SSS
   .mode("overwrite")
   .save("Worktime2.xlsx")
 ```
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ val df = sqlContext.read
     .option("addColorColumns", "true") // Optional, default: false
     .option("startColumn", 0) // Optional, default: 0
     .option("endColumn", 99) // Optional, default: Int.MaxValue
+    .option("timestampFormat", "hh:mm:ss:SSS") // Optional, default: yyyy-mm-dd hh:mm:ss[.fffffffff]
     .schema(myCustomSchema) // Optional, default: Either inferred schema, or all columns are Strings
     .load("Worktime.xlsx")
 ```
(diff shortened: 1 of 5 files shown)


nightscape:
Hey @davidw76,

this looks like a nice thing to add. In the long run, one would probably even want to somehow specify a date format per column, but this is definitely a step in the right direction.
Could you update the `README.md` accordingly and try adapting [the tests](https://github.com/crealytics/spark-excel/blob/master/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala)? You could e.g. add a `aDateFormattedAsString: String` field below [this line](https://github.com/crealytics/spark-excel/blob/master/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala#L26) and change [the code here](https://github.com/crealytics/spark-excel/blob/master/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala#L84) to something like
```scala
val parseTimestampUdf = udf(timestampString: String => parseTimestampWithFormat(timestampString, format))
val expected = spark.createDataset(rows).toDF.withColumn("aDateFormattedAsString", parseTimestampUdf($"aDateFormattedAsString"))
```

nightscape:
Also, should the same formatter be applied when writing Excel files? If so, [this would be the right place](https://github.com/crealytics/spark-excel/blob/master/src/main/scala/com/crealytics/spark/excel/ExcelFileSaver.scala#L31).
That might even simplify the tests a little.

davidw76:
For writing files, the line you linked to is for type Date. I think that timestampFormat should only apply to type Timestamp.
I could add two options to save.. dateFormat and timestampFormat. Would that be OK?

davidw76:
Hi Martin, I have made the following changes...

- Added a _timestampFormat_ option to ExcelFileSaver. Timestamps are written to Excel as strings with the given formatting (or default formatting if none given). I did not add a _dateFormat_ option because dates are exported to Excel as numbers with the Excel format string being used only for display purposes.
- Updated the integration test to include a timestamp column in the data. This is populated with random values, written out to the Excel file using the default format, and then read back in as a Timestamp column (again using the default format) and compared with the original. I also verified that it worked with custom formats but there is no test for that currently. It would probably be better to add a TimestampSuite with unit tests to cover that.
- Updated the readme with the _timestampFormat_ options.

PR #25, opened 2017-09-07
added executor-base data source.

dmytroDragan:
Added as another source  executor-base data source. Extract common excel processing logic.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ project/target/
 .ensime*
 *.swp
 .idea
+*.log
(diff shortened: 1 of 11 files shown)


nightscape:
Hi @3Dragan,

thanks for creating this PR! May I ask you to nonetheless to minimize the changeset (and rebase on the latest master)?
It's a hard to review the actual logic change as you moved a lot of code and changed the formatting.
I'm not opposed to doing refactorings, but would like to review them separately.

Thanks!

3Dragan:
Hi, I have reset origin parts of upstream.

nightscape:
@3Dragan are you still interested in this?
If so, would you be so kind to create a version of this PR with a more minimal changeset?
It's really hard to review like that because we'd almost have to go through every file line by line and check if you changed something there or just moved the code to another location...

nightscape:
The code has changed quite a bit in the mean time. I've got a branch which switches to Excel being a `FileFormat` which is how Spark's internal CSV reader is implemented as well.

PR #26, opened 2017-09-10
write timestamp as number with format option and add format option fo…

davidw76:
I had second thoughts about saving Timestamps as strings. I think it is more useful to save them as numbers so that they can be used in Excel calculations.

This pull request does the following...
- Timestamps saved to Excel as numbers with optional formatting string. Default format is `yyyy-mm-dd hh:mm:ss.000`
- Added optional formatting string for dates. Default is unchanged as `yy-m-d h:mm`

One potentially confusing thing is that when reading a file, `timestampFormat `is in Java SimpleDateFormat style for parsing. 
But when writing a file, `dateFormat `and `timestampFormat `are in the Excel style for display formatting. Maybe we should add a note about this to README.md.

With this change, it is still possible to save the Timestamp column and reload it as the same type, provided the same formatting is used in both save and read. No changes to the tests were necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ val df = sqlContext.read
     .option("addColorColumns", "true") // Optional, default: false
     .option("startColumn", 0) // Optional, default: 0
     .option("endColumn", 99) // Optional, default: Int.MaxValue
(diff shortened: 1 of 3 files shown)


davidw76:
no, nothing else to include in this one thanks

PR #28, opened 2017-09-20
Detect date formatted columns when inferring schema

davidw76:
Currently, spark-excel does not try to infer the type of date/time columns.

With this change, if the column is formatted with date and/or time portions in Excel, then spark-excel will infer the column type as a `Timestamp`. 
Inferring as `Timestamp` is more useful than inferring it as a `Date `because the spark `java.sql.Date` type hides any time component that might be present (hours, minutes, seconds, milliseconds).
The check for data/time formatting is done using the POI function.. `DateUtil.isCellDateFormatted`.

This PR also includes the change discussed [here](https://github.com/crealytics/spark-excel/pull/27).

### DIFF
--- a/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
+++ b/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
@@ -18,16 +18,16 @@ import org.apache.spark.sql.functions.lit
 object IntegrationSuite {
 
   case class ExampleData(
-    aBoolean: Boolean,
-    aByte: Byte,
-    aShort: Short,
-    anInt: Int,
-    aLong: Long,
-    aDouble: Double,
-    aString: String,
-    aTimestamp: java.sql.Timestamp,
-    aDate: java.sql.Date
-  )
+                          aBoolean: Boolean,
+                          aByte: Byte,
+                          aShort: Short,
+                          anInt: Int,
+                          aLong: Long,
+                          aDouble: Double,
+                          aString: String,
+                          aTimestamp: java.sql.Timestamp,
+                          aDate: java.sql.Date
+                        )
 
   val exampleDataSchema = ScalaReflection.schemaFor[ExampleData].dataType.asInstanceOf[StructType]
 
@@ -67,7 +67,7 @@ class IntegrationSuite extends FunSuite with PropertyChecks with DataFrameSuiteB
   val PackageName = "com.crealytics.spark.excel"
   val sheetName = "test sheet"
 
-  def writeThenRead(df: DataFrame): DataFrame = {
+  def writeThenRead(df: DataFrame, schema: Option[StructType] = Some(exampleDataSchema)): DataFrame = {
     val fileName = File.createTempFile("spark_excel_test_", ".xlsx").getAbsolutePath
 
     df.write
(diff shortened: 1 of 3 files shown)


nightscape:
Hi @davidw76,

thanks for keeping on improving this plugin! :smiley:
Is the new functionality already covered by existing tests?
Could you rebase on master and squash the commits?

davidw76:
The changes effect the case where the schema is inferred instead of given as a parameter. There don't seem to be any tests covering this. I will add one.

davidw76:
I have added a unit test for infer schema and squashed the commit history.

src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
```diff
@@ -18,16 +18,16 @@ import org.apache.spark.sql.functions.lit
...
-    aByte: Byte,
-    aShort: Short,
-    anInt: Int,
-    aLong: Long,
-    aDouble: Double,
-    aString: String,
-    aTimestamp: java.sql.Timestamp,
-    aDate: java.sql.Date
-  )
+                          aBoolean: Boolean,
```
nightscape: This indentation style leads to unnecessary git diffs when `ExampleData` gets renamed to something shorter or longer. Would you be so kind to change it back?

I will add a scalafmt configuration in a few minutes, then all formatting questions should be automated :wink:

src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
```diff
@@ -111,4 +114,21 @@ class IntegrationSuite extends FunSuite with PropertyChecks with DataFrameSuiteB
       assertDataFrameEquals(expectedWithEmptyStr, writeThenRead(expectedWithNull))
     }
   }
+
+  test("infers schema correctly") {
+    forAll(rowsGen, MinSuccessful(20)) { rows =>
+      val df = spark.createDataset(rows).toDF
+      val inferred = writeThenRead(df, schema = None)
+
+      assert(if (df.count() > 0) BooleanType else StringType, inferred.schema.fields(0).dataType)
```
nightscape: You could rewrite this to something like (untested):
```scala
val expectedDataTypes =
  if (df.count() == 0) {
    // Without actual data, we assume everything is a StringType
    exampleDataSchema.fields.map(_ => StringType)
  } else {
    exampleDataSchema.fields.map(_.dataType)
  }
assert(inferred.schema.fields.map(_.dataType), expectedDataTypes)

```

src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
davidw76: I will fix it.. it didn't show up in my merge tool sorry

src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
davidw76: OK I will tidy it up

davidw76: We cannot use the types from exampleDataSchema because the inferred types will not match exactly. For example, the byte, short, int and long columns will all be inferred as DoubleType.

nightscape: Ok, then maybe just make it a `Seq(DoubleType, DoubleType, ..., TimestampType)`

davidw76: Yep

davidw76:
I restored the formatting and tidied up the unit test

nightscape:
Thanks again for your contributions!
I just released 0.9.4 with your changes. It should appear on the Maven repos within a few hours.

davidw76:
You're welcome.. the library is very useful thanks

Issue #30, opened 2017-10-16
formulas in sheets do not appear to be supported


habelson:
The values read from cells in a sheet that are of type FORMULA, are evaluated as the formula itself rather than it's cached value.

Additionally the infer schema option fails on cells of type FORMULA.

OK to submit a PR for this?

nightscape:
Definitely :+1:

nightscape:
Fixed by #31

PR #31, opened 2017-10-18
Added support for FORMULA cells when typing to String or Numeric types

habelson:
Pull Request to support issue #30 . This patch will work in most cases where cached values exist in the saved worksheet. I was not able to add support for typing other than String or Numeric, perhaps someone else can embellish.

### DIFF
--- a/src/main/scala/com/crealytics/spark/excel/ExcelRelation.scala
+++ b/src/main/scala/com/crealytics/spark/excel/ExcelRelation.scala
@@ -104,12 +104,28 @@ case class ExcelRelation(
     }
 
     val dataFormatter = new DataFormatter()
-    lazy val stringValue = dataFormatter.formatCellValue(cell)
+    lazy val stringValue =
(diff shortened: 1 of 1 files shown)


nightscape:
Out of interest: Could you also [evaluate the cells](https://poi.apache.org/spreadsheet/eval.html) if no cached value exists?

habelson:
@nightscape I do not think that apache poi has support for that. That would require duplicating the excel formulas engine in the library which i think would be far outside the scope of this plugin.

nightscape:
Yeah, that would probably be the best option.
Would you mind adding that?
If not, I'm also OK with merging it as is.

litzebauer:
@nightscape Any chance you can merge this change in?

nightscape:
I just rebased this PR and released it as 0.9.9. It will probably take a few hours until it is available via Sonatype.
Could you give it a whirl and report back here in case there are problems?

Issue #38, opened 2017-11-14
infer schema only reading first ten rows?


davidw76:
With version 0.9.6, when I do not provide a schema and set `inferSchema` to true, the inferred schema seems to be based on the first ten rows only.
This means that if a column contains only numbers in the first ten rows, but contains text (that cannot be parsed as a number) in later rows, then there will be a crash when trying to parse the text as a number.
If there is non-numeric text in the first ten rows of data then the column is correctly inferred as a string type and there is no problem.
I am pretty sure that in previous versions, the schema was inferred from all rows and this problem never occurred.
I have `maxRowsInMemory` set to 20
The problem can be seen with the [online retail dataset](http://archive.ics.uci.edu/ml/machine-learning-databases/00352/)

davidw76:
I guess it is because excerptSize is set to 10 and inferSchema is done on the excerpt which contains only ten rows?
Is this intentional?

davidw76:
I guess it is intentional to infer the schema from the first ten rows. I would like to make the following changes..

1. Make `excerptSize` a read option. The default can be ten. This option would only apply when `inferSchema` is true
2. Make parsing more permissive so that if a number is expected but it's a string value that cannot be parsed as a number, set it to `NaN`

nightscape:
Hi @davidw76,

that's a good idea! I intended to make the `excerptSize` configurable, but forgot about it in the end...
So, if you want to create a PR, I'd be happy to review and release it.

PR #39, opened 2017-11-15
#38 excerptSize option

davidw76:
This is the fix for #38 
I have made `excerptSize` a read option with a default of `10`
Also made parsing a bit more permissive so that if `excerptSize` is too small to catch a non-numeric in a mostly numeric column, the numeric value will be set to NaN instead of throwing an exception.


nightscape:
Merged and released as `0.9.7`.
Thank you!

Issue #19, opened 2017-09-04
Out of memory exception when reading big xlsx file


aprasa02:
I am reading big xlsx file of 100mb with 28 sheets(10000 rows per sheet) and creating a single dataframe out of it . I am facing out of memory exception when running on cluster mode .My code looks like this.

def buildDataframe(spark: SparkSession, filePath: String, requiresHeader: Boolean): DataFrame = {

    var excelDF: DataFrame = null
    var flag: Boolean = false
// getting inputstream from s3 for all sheet names   
 val s3FilePath = filePath.replaceAll(filePath.substring(0, filePath.indexOf("//") + 2), "")
    val s3Bucket = s3FilePath.substring(0, s3FilePath.indexOf("/"))
    val s3key = s3FilePath.substring(s3FilePath.indexOf("/") + 1, s3FilePath.length())
    val s3Client: AmazonS3 = new AmazonS3Client();
    val s3object: S3Object = s3Client.getObject(new GetObjectRequest(
      s3Bucket, s3key));
    val inputStream: java.io.InputStream = s3object.getObjectContent()
    val workbook = new XSSFWorkbook(inputStream)
    var sno = workbook.getNumberOfSheets
    var sheetName: Array[String] = new Array[String](sno)
    for ((sheet, i) <- workbook.zipWithIndex) {
      val initialDF = spark.read.format("com.crealytics.spark.excel")
        .option("useHeader", requiresHeader)
        .option("treatEmptyValuesAsNulls", "true")
        .option("inferSchema", true)
        .option("addColorColumns", false)
        .option("sheetName", sheet.getSheetName)
        .option("path", filePath)
        .load()

      if (flag == false) {
        excelDF = initialDF
        flag = true
      } else {
        excelDF = excelDF.union(initialDF)
      }

    }

    return excelDF
  }

Any help is much appreciated.

gregosaurus:
Hello @aprasa02 !

What are the value set for `--driver-memory`, `--executor-memory` and `--executor-cores` when submitting your Spark Job ?

How many partitions does your file contains ?

Usually 100mb are not a problem for Spark.

Best,

Gregory.

aprasa02:
[...]

PR #41, opened 2017-11-16
shade dependency on jackson

davidw76:
this avoids an incompatibility with the version of jackson packaged with spark
see comments in #19


nightscape:
LGTM!

nightscape:
I published `0.9.8` with this change. Should land in the repos within the next hours.

PR #57, opened 2018-02-12
Support for reading/writing from/to directory

kholodilov:
Support for writing/reading files, when path specified to load/save methods is a directory.

Introduced new parameters:
- writeToFile - optional string parameter for writer, not set by default. Allows to specify filename to write excel file inside "path" directory. Also causes writer to create empty "_SUCCESS" flag inside "path".
- readFromFile - optional string parameter for reader, not set by default. Allows to specify filename to read excel file inside "path" directory.

### DIFF
--- a/src/main/scala/com/crealytics/spark/excel/DefaultSource.scala
+++ b/src/main/scala/com/crealytics/spark/excel/DefaultSource.scala
@@ -23,6 +23,7 @@ class DefaultSource extends RelationProvider with SchemaRelationProvider with Cr
   ): ExcelRelation = {
     ExcelRelation(
       location = checkParameter(parameters, "path"),
+      readFromFile = parameters.get("readFromFile"),
       sheetName = parameters.get("sheetName"),
       useHeader = checkParameter(parameters, "useHeader").toBoolean,
       treatEmptyValuesAsNulls = parameters.get("treatEmptyValuesAsNulls").fold(true)(_.toBoolean),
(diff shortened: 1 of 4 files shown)


nightscape:
Hi @kholodilov,

can you give a little more background and what you need this for?
Is it just for creating the directory or is there more to it?

kholodilov:
This is standard approach of all default Spark datasources. Main benefit is that writes become transactional (thanks to _SUCCESS flag), so if downstream process needs to read data written by your process, it won't read half-written data (will wait for flag).
Creating the directory automatically also adds convenience.

geoHeil:
will this be merged?

nightscape:
The reason why I'm a little hesitant to merge this as is, is that usually the creation of `_SUCCESS` files is done by Hadoop along with `.crc` files for every of the `part-...` files.
I'm not sure if it's a good idea to reproduce this behavior only half-way or if one should actually call the appropriate Hadoop methods.
Also, it would be nice if the tests reused more of the existing code.

src/main/scala/com/crealytics/spark/excel/DefaultSource.scala
```diff
@@ -44,6 +45,7 @@ class DefaultSource extends RelationProvider with SchemaRelationProvider with Cr
     data: DataFrame
   ): BaseRelation = {
     val path = checkParameter(parameters, "path")
+    val writeToFile = parameters.get("writeToFile")
```
nightscape: Can we unify `readFromFile` and `writeToFile` to something like `fileLocation` or `filePath`?

nightscape:
Superseded by the (still unfinished) https://github.com/crealytics/spark-excel/pull/210

PR #58, opened 2018-02-12
Support for writing/reading files with legal disclaimer

kholodilov:
Support for writing/reading files with legal disclaimer (like "Confidential. All rights reserved (c) Company Name") in first row.

Introduced new parameters:
- legalDisclaimer - optional string parameter for writer, not set by default. Allows to write arbitrary string in the first cell of the first row of workbook.
- skipFirstRows - optional integer parameter for reader, not set by default. Allows to read data from files with arbitrary strings in first N rows (= skip this rows).

### DIFF
--- a/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
+++ b/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
@@ -1,16 +1,17 @@
 package com.crealytics.spark.excel
 
 import java.io.File
+import java.lang.Math.sqrt
 import java.sql.Timestamp
 
 import org.scalacheck.{Arbitrary, Gen, Shrink}
-import Arbitrary.{arbLong => _, arbString => _, arbBigDecimal => _, _}
+import Arbitrary.{arbBigDecimal => _, arbLong => _, arbString => _, _}
 import org.scalacheck.ScalacheckShapeless._
-import org.scalatest.FunSpec
+import org.scalatest.{FunSpec, Matchers}
 import org.scalatest.prop.PropertyChecks
 import com.holdenkarau.spark.testing.DataFrameSuiteBase
 import org.apache.hadoop.fs.Path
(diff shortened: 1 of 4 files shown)


src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala
```diff
@@ -195,6 +196,48 @@ class IntegrationSuite extends FunSpec with PropertyChecks with DataFrameSuiteBa
         }
       }
 
+      it("writes data with legal disclaimer") {
+        val fileName = File.createTempFile("spark_excel_test_", ".xlsx").getAbsolutePath
+
+        val df =
+          Seq((1, "1", "2", "3"), (2, "4", "5", "6"), (3, "7", "8", "9")).toDF("id", "column1", "column2", "column3")
+
+        df.write
```
nightscape: Can you try to reuse more of the test infrastructure from the other tests? In order to do so, you could pass an additional argument Map which will get passed to the `option` or `options` method.
The test right now e.g. doesn't test the streaming version (`maxRowsInMemory = Some(20)`) which is working quite differently under the hood.

nightscape:
I think `legalDisclaimer` is just an edge case of a more generic concern to write data that is before headers and data.
Can you rename `legalDisclaimer` to something more generic?
In a later step (or if you want, now) we could make sure that `\n` gets transformed into a new row and `\t` gets transformed into a new cell in a row.

nightscape:
Closing this in favor of https://github.com/crealytics/spark-excel/pull/72

nightscape:
Released as `0.9.16`.

nightscape:
Hi @kholodilov,

I'm just implementing a feature to write to existing files at arbitrary locations (probably also with an option to use a read-only template file and write to a different location).
Would this cater for your use case? If so, I'd like to remove the `preHeader` option in favour of the new, more powerful mechanism.

PR #61, opened 2018-04-08
add Float type in toCell

sadhen:
```
java.lang.RuntimeException: scala.MatchError: 0.0 (of class java.lang.Float)
at com.crealytics.spark.excel.ExcelFileSaver.toCell(ExcelFileSaver.scala:44)
...
```

Don't know why Float is left out. I encountered the above error using Spark 2.1.1 and spark-excel master.


nightscape:
Probably just because we never used `Float`s, but rather `Double`s.
It would be nice if you could enhance the test here:
https://github.com/crealytics/spark-excel/blob/master/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala#L26

sadhen:
@nightscape  Updated, please review

nightscape:
Published as 0.9.15

PR #67, opened 2018-05-15
Io error fix

sterkh66:


src/main/scala/com/crealytics/spark/excel/DefaultSource.scala
```diff
@@ -44,10 +46,12 @@ class DefaultSource extends RelationProvider with SchemaRelationProvider with Cr
     data: DataFrame
   ): BaseRelation = {
     val path = checkParameter(parameters, "path")
+    val writeToFile = parameters.get("writeToFile")
```
nightscape: Can we unify `readFromFile` and `writeToFile` to something like `fileLocation` or `filePath`?

src/main/scala/com/crealytics/spark/excel/DefaultSource.scala
```diff
@@ -44,10 +46,12 @@ class DefaultSource extends RelationProvider with SchemaRelationProvider with Cr
...
     val sheetName = parameters.getOrElse("sheetName", "Sheet1")
     val useHeader = checkParameter(parameters, "useHeader").toBoolean
     val dateFormat = parameters.getOrElse("dateFormat", ExcelFileSaver.DEFAULT_DATE_FORMAT)
     val timestampFormat = parameters.getOrElse("timestampFormat", ExcelFileSaver.DEFAULT_TIMESTAMP_FORMAT)
+    val legalDisclaimer = parameters.get("legalDisclaimer")
```
nightscape: I think `legalDisclaimer` is just an edge case of a more generic concern to write data that is before headers and data.
Can you rename `legalDisclaimer` to something more generic?
In a later step (or if you want, now) we could make sure that `\n` gets transformed into a new row and `\t` gets transformed into a new cell in a row.

PR #72, opened 2018-07-06
preHeader support

nightscape:
This is a slight change to https://github.com/crealytics/spark-excel/pull/58
The parameter for adding stuff above the actual headers is now called `preHeader` because it could contain other stuff like summaries, comments, ...
If `preHeader` contains line-breaks, these get transformed into new rows.
The tests are now also randomized via ScalaCheck (which immediately uncovered a problem with empty `preHeader = Some("")` leading to a `?` being written in some cases, a problem I chose to ignore for the time being).
@kholodilov could you have a look if there's any regressions or any improvements you'd like to add?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,13 +2,13 @@ name := "spark-excel"
 
 organization := "com.crealytics"
 
-crossScalaVersions := Seq("2.11.11", "2.10.6")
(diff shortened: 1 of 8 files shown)


nightscape:
Released as `0.9.16`.

PR #80, opened 2018-09-08
Upgrade to POI 4.0.0

pjfanning:
I have forked the xlsx-streamer to support poi 4.0.0 but the APIs are basically the same.
The spoiwo library will probably also need to be upgraded when they publish a poi 4.0.0 compatible jar (they've already merged a change to support this).


nightscape:
Hi @pjfanning,

nice job, thanks for the contribution!
I would like to wait for your PR against xlsx-streamer to be merged, so that we don't switch dependencies and imports back and forth. Is that ok for you?

nightscape:
Rebased and merged in f775c72106506e4ddcec65a4913c14f4e7025f70

nightscape:
Released as 0.10.0. Will take a few hours to hit Maven Central though.

PR #84, opened 2018-09-26
 Read password protected excel files

### LINKED ISSUE #83: Read password protected excel files
Apache poi can read password protected  HSSF and XSSF files.
It would be nice to add this option to spark-excel.

Use case:
Spread sheet that contains personal information might be required by business to be password protected.

PanAeon:
resolve  #83
I believe this is quite streightforward change, but:
* This change requires unlimited strength JCE to work, including for unit tests
* I haven't add encrypted workbook write support. But I believe it is possible, see [the TestWorkbookProtection testcase](https://bz.apache.org/bugzilla/attachment.cgi?id=34349)


nightscape:
Great job! Merging and releasing now 👍

PR #91, opened 2018-10-17
Changed sheetName param in the Readme.md to optional. 

PanAeon:
Added test to verify that it is working as optional.
I looked in the code and when sheetName is not specified then spark will read the first sheet in the workbook. I think the confusion in the readme.md rose because for the writes sheetname is obviously required (in a sense that it cannot be null).

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ import org.apache.spark.sql._
 val spark: SparkSession = ???
 val df = spark.read
     .format("com.crealytics.spark.excel")
-    .option("sheetName", "Daily") // Required
+    .option("sheetName", "Daily") // Optional, if not set reads the first sheet in the spreadsheet
     .option("useHeader", "true") // Required
     .option("treatEmptyValuesAsNulls", "false") // Optional, default: true
     .option("inferSchema", "false") // Optional, default: false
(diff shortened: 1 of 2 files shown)


nightscape:
Hi @PanAeon, I'm currently rewriting that entire logic.
I'm planning to replace `sheetName` and friends by a `dataAddress` option which accepts things like

- `A1:C25` (uses the first sheet)
- `MySheet!A1:C25`
- `MyTable[#All]`
- `MyTable[[ColOne]:[ColFour]]`

I was thinking about leaving `sheetName` and adding `tableName`, `tableColumns`, etc. but that would lead to a lot of options which might not be compatible with each other (e.g. `sheetName` and `tableName`, `cellRange` and `tableColumns`).
But I'm still open to other ideas.

PanAeon:
Well, I think it's important that people could still read the first sheet without specifying it's name. For one-sheet files there should be no confusion, and programmer might not know beforehand what that sheet would be called. Sometimes there're no choice whatsoever - the needed sheet might be called differently in a set of spreadsheets.
To that extends I think it would be also useful to have a way to specify sheet number instead of name of the sheet.

nightscape:
I would have seen the first sheet as a special case, so if the user doesn't provide a `dataAddress` then the entire first sheet will be read.
If one provides only a cell range like `A1:C35` it will also be this range of the first sheet.
Do you think there's also a case where one wants to take the second sheet without knowing its name?

nightscape:
The `dataAddress` option is in place now.

PR #229, opened 2020-03-31
Fix timestamp formula supporting

ylevin:
Case when timestamp cell contains formula missed


nightscape:
Looking good, let's see if the new PR Github Action triggers from my comment 🤞

nightscape:
Nope, doesn't seem so...

nightscape:
@ylevin this should become a new snapshot version once the corresponding Github Actions finish:
https://github.com/crealytics/spark-excel/actions/runs/68628862

Issue #223, opened 2020-03-20
maxRowsInMemory can cause rows to be dropped


jamthoma:
I have a relatively large xls that causes memory issues, so I tried to use "maxRowsInMemory".
Unfortunately, this caused most of the rows to get dropped.
I puled the source code and uncommented the exception in ExcelRelation.scala (line 73).
The exception below was observed.

Basically, version 2.1.0 of xlsx-streamer does not implement "getSheet", which is used by POI DataFormatter.

The latest code in xlsx-streamer has implemented the "getSheet" API, so once it's released this issue will hopefully be resolved by increasing the xlsx-streamer dependency version.

> shadeio.monitorjbl.xlsx.exceptions.NotSupportedException
        at shadeio.monitorjbl.xlsx.impl.StreamingCell.getSheet(StreamingCell.java:330)
        at shadeio.poi.ss.usermodel.DataFormatter.isDate1904(DataFormatter.java:313)
        at shadeio.poi.ss.usermodel.DataFormatter.getFormat(DataFormatter.java:309)
        at shadeio.poi.ss.usermodel.DataFormatter.getFormattedNumberString(DataFormatter.java:858)
        at shadeio.poi.ss.usermodel.DataFormatter.formatCellValue(DataFormatter.java:1011)
        at shadeio.poi.ss.usermodel.DataFormatter.formatCellValue(DataFormatter.java:961)
        at shadeio.poi.ss.usermodel.DataFormatter.formatCellValue(DataFormatter.java:940)
        at com.crealytics.spark.excel.HeaderDataColumn.stringValue$lzycompute$1(DataColumn.scala:43)
        at com.crealytics.spark.excel.HeaderDataColumn.stringValue$1(DataColumn.scala:33)
        at com.crealytics.spark.excel.HeaderDataColumn.extractValue(DataColumn.scala:79)
        at com.crealytics.spark.excel.DataColumn$class.apply(DataColumn.scala:16)
        at com.crealytics.spark.excel.HeaderDataColumn.apply(DataColumn.scala:19)
        at com.crealytics.spark.excel.HeaderDataColumn.apply(DataColumn.scala:19)
'

nightscape:
Hi @jamthoma, nice find!
I think @scala-steward should let us know once the new version of `xlsx-streamer` hits Maven :+1:

karamveer-scala:
@nightscape  

[...]

PR #246, opened 2020-05-11
POI 4.1.2 and switch to excel-streaming-reader fork

pjfanning:
https://github.com/crealytics/spark-excel/issues/223

* I didn't deliberately reformat the scala classes - seem to have had scalafmt run on them by a build
* one extra change I made was on the bufferSize - this param is not really associated with the maxRowsInMemory setting so I just changed it to a sensible default value


nightscape:
Perfect, thanks a lot @pjfanning !

PR #253, opened 2020-05-20
Publish on new version tag

EnverOsmanov:
As I can see in commit history, there are no commits for release. Only a tag is being created. So "publish" step should happen for tags.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
         run: sbt ++${{ matrix.scala }} scalastyle "test:scalastyle"
   publish:
     needs: build
-    if: contains(github.ref, 'refs/tags/v')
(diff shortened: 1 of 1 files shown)


EnverOsmanov:
@nightscape, I noticed that during pull requests github actions are being triggered twice- for push and for pull request. It would be better to have only for pull_requests. What do you think?

.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: sbt ++${{ matrix.scala }} scalastyle "test:scalastyle"
   publish:
     needs: build
-    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
+    if: contains(github.ref, 'refs/tags/v')
```
nightscape, line 26: Removing the `refs/heads/master` would disable the snapshot builds (which I find rather nice). Could you change this to trigger for both?

EnverOsmanov: Done.

PR #281, opened 2020-08-11
Upgrade to Spark 3.0

Fokko:
We're upgrading to Spark 3.0 and we'd like to use the Excel reader as well :)

Ran the tests locally and they passed.


Fokko:
Hmm, I see that the CI is already running against 3.0

PR #294, opened 2020-09-23
fix example code the map object is making error

gwdgithubnom:
change to Map("path" -> "Worktime.xlsx")


nightscape (approved):

nightscape:
Good catch, thanks!

Issue #321, opened 2020-11-26
Issue with reading a long number > 11 digit


waiyan1612:
## Expected Behavior
I have a simple xlsx file with a column that can have numbers more than 11 digits.
```
+------------+
|id          |
+------------+
|12345678901 |
|123456789012|
|123456789-0 |
+------------+
```
I want to create a DataFrame with id as a `StringType` and read all the values as they are (i.e. Not converted to a scientific notation etc)

## Current Behavior
I have tried the following options.
1. custom schema with `id`=`StringType`
2. `inferSchema=True`
3. `inferSchema=False`
4. default options.

Every option gives me the same output.
```
+-----------+
|id         |
+-----------+
|12345678901|
|1.23457E+11|
|123456789-0|
+-----------+

root
 |-- id: string (nullable = true)
```
The issue lies in the second row. I am unable to get `123456789012`. Parsing `1.23457E+11` to `Long` does not help because I have already lost the precision. 

I also noticed that this issue does not occur if this column is [formatted](https://docs.microsoft.com/en-us/office/troubleshoot/excel/long-numbers-incorrectly-in-excel) as `Text` instead of `General` in Excel. Unfortunately, I do not have control over this.

## Steps to Reproduce (for bugs)
1. Download the attached excel file.
[long_number.xlsx](https://github.com/crealytics/spark-excel/files/5601631/long_number.xlsx)

2. Read the downloaded example file
    ```scala
    val spark: SparkSession = SparkSession.builder()
      .appName("test-excel")
      .master("local[*]")
      .getOrCreate()

    val path = "long_number.xlsx"
    val columns = spark.read.format("com.crealytics.spark.excel")
      .option("header", true)
      .load(path).columns
    val allStringSchema = StructType(columns.map(x => StructField(x, StringType, true)))
    println(allStringSchema)

    println("Using customSchema", allStringSchema)
    val df = spark.read.format("com.crealytics.spark.excel")
      .option("header", true)
      .schema(allStringSchema)
      .load(path)
    df.show(false)
    df.printSchema()

    println("Using inferSchema=true")
    val df2 = spark.read.format("com.crealytics.spark.excel")
      .option("header", true)
      .option("inferSchema", true)
      .load(path)
    df2.show(false)
    df2.printSchema()

    println("Using inferSchema=false")
    val df3 = spark.read.format("com.crealytics.spark.excel")
      .option("header", true)
      .option("inferSchema", false)
      .load(path)
    df3.show(false)
[...]

PR #329, opened 2020-12-07
Do not use rounding and scientific notations for long numbers in Excel's General format

waiyan1612:
## Description
This PR addresses https://github.com/crealytics/spark-excel/issues/321.
- `PlainNumberFormat` is extended from `Format` similar to `ExcelGeneralNumberFormat`. Rather than formatting the decimals, it uses `BigDecimal.toPlainString()` implementation. 
- This format is added to `DataFormatter` similar to how `SSNFormat` [is being added](https://github.com/apache/poi/blob/REL_4_1_2/src/java/org/apache/poi/ss/usermodel/DataFormatter.java#L1157-L1159). Instead of adding as a new format, we overwrite the existing `General` and `@` formats whose previous value is `ExcelGeneralNumberFormat`.
- Tested with decimals, strings, dates and formulas. See the test case for details.


src/main/scala/com/crealytics/spark/excel/DataColumn.scala
```diff
@@ -30,6 +30,12 @@ class HeaderDataColumn(
     }
 
     lazy val dataFormatter = new DataFormatter()
+    // Overwrite ExcelGeneralNumberFormat with custom PlainNumberFormat.
+    // See https://github.com/crealytics/spark-excel/issues/321
+    lazy val plainNumberFormat = PlainNumberFormat.getInstance()
+    dataFormatter.addFormat("General", plainNumberFormat)
+    dataFormatter.addFormat("@", plainNumberFormat)
```
waiyan1612, line 37: Do we want to add a config to enable / disable this behaviour (similar to `treatEmptyValuesAsNulls` in this class?

nightscape: Depending on whether it is possible that this breaks somebody's workflow, it might be a good idea :wink:

src/test/scala/com/crealytics/spark/excel/PlainNumberReadSuite.scala
```diff
@@ -0,0 +1,68 @@
...
+  val expectedDataNonInferSchema: util.List[Row] = List(
+    Row("12345678901", "12345678901-123", "12/1/20"),
+    Row("123456789012", "123456789012", "0.01"),
+    Row("-0.12345678901", "0.05", "0h 14m"),
+    Row(null, null, null),
+    Row("-1/0", "abc.def", null)
+  ).asJava
+}
+
+class PlainNumberReadSuite extends AnyFunSpec with DataFrameSuiteBase with Matchers {
```
nightscape, line 46: Great!! Thanks for adding tests as well!

nightscape (commented):
I'm thinking about whether we convert `PlainNumberFormat` to Scala in order to stick to one language.
The code would be the following:
```scala
package com.crealytics.spark.excel

import java.math.BigDecimal
import java.text.FieldPosition
import java.text.Format
import java.text.ParsePosition

/**
  * A format that formats a double as a plain string without rounding and scientific notation.
  * All other operations are unsupported.
  * @see <a href="https://github.com/apache/poi/blob/REL_4_1_2/src/java/org/apache/poi/ss/usermodel/ExcelGeneralNumberFormat.java">
  * ExcelGeneralNumberFormat</a> and <a href="https://github.com/apache/poi/blob/REL_4_1_2/src/java/org/apache/poi/ss/usermodel/DataFormatter.java#L1171">
  * SSNFormat</a> from Apache POI.
  *
  */
object PlainNumberFormat extends Format {

  override def format(number: AnyRef, toAppendTo: StringBuffer, pos: FieldPosition): StringBuffer =
    toAppendTo.append(new BigDecimal(number.toString).toPlainString())

  override def parseObject(source: String, pos: ParsePosition): AnyRef =
    throw new UnsupportedOperationException()
}
```

waiyan1612:
Thanks for the feedbacks. Sticking to scala for `PlainNumberFormat` and also introduced `usePlainNumberFormat` config.

nightscape (approved):
Great job! Will merge, you should get a build in a few minutes.

nightscape:
0.13.3+9-bea39dbd-SNAPSHOT it is.
@waiyan1612 would you be interested in becoming a contributor and getting additional permissions?

waiyan1612:
@nightscape Thank you for the offer! I am still new to this but I am interested in helping out with the project.

> 0.13.3+9-bea39dbd-SNAPSHOT it is.

Do you mean https://oss.sonatype.org/content/repositories/public/com/crealytics/spark-excel_2.12/0.13.3+9-bea39dbd-SNAPSHOT/? I decompiled this jar but cannot see the new classes inside. I can see them when I build the master branch locally.

nightscape:
@waiyan1612 looks like I had messed up the automatic publishing of pushes to the `main` (previously `master`) branch.
This should be fixed now and I created a `0.13.6` release.

waiyan1612:
Thank you! I have tested 0.13.6 and can verify the fix. Let me close the original issue.

nightscape:
Thank you for the contribution :smiley:
I've given you triage rights to the repo, if you want/dare to get more let me know and I'll give you more permissions :wink:

Issue #371, opened 2021-05-07
Unable to Partition while saving as Excel file


KashyapCKotak:
I am using the below version of spark-excel:
```
groupId: com.crealytics
artifactId: spark-excel_2.11
version: 0.13.1
```

I have a Dataset similar to the below structure:
```
col_A      col_B        date
  1          5       2021-04-14
  2          7       2021-04-14
  3          5       2021-04-14
  4          9       2021-04-14
```

I am trying to use the below code in Spark Java to write the dataaset to a file in HDFS.

```
Dataset<Row> outputDataset; // This is a valid dataset and works flawlessly when written to csv
/*
   some code which sets the outputDataset
*/
outputDataset
    .repartition(1)
    .write()
    .partitionBy("date")
    .format("com.crealytics.spark.excel")
    .option("header", "true")
    .save("/saveLoc/sales");
```

Normal Working Case:

When I pass use `.format("csv")`, the above code creates a folder with the name `date=2021-04-14` in the path `/saveLoc/sales` that is passed in `.save()` which is exactly as expected. The full path of the end file is `/saveLoc/sales/date=2021-04-14/someFileName.csv`. Also, the column `date` is removed from the file since it was partitioned on.

What I need to do:

However, when I use `.format("com.crealytics.spark.excel")`, it just creates a plain file called `sales` in the folder `saveLoc` and doesn't remove the partitioned(date) column from the end file. Does that mean it isn't partitioning on the column "date"? Full path of the file created is `/saveLoc/sales`. Please note that it overrides the folder "sales" with a file sales.

How can I make it parition when writing in excel? In other words, how can I make it behave exactly as it did in case of csv?

quanghgx:
Hi @KashyapCKotak 
This partitioning write should be well supported by now. Please take a try again, and use .format("excel") instead.
I am going to resolve this ticket, feel free to reopen it.
Sincerely,

Issue #285, opened 2020-08-31
Empty rows are automatically removed. Is it possible to not have that behavior?


Bryce21:
## Expected Behavior
I'm using an excel sheet that contains empty rows. When I read that into a dataframe it is losing the empty rows. My data is very delicate and needs those empty rows for downstream processes.

## Current Behavior
I'm losing the empty rows.

## Steps to Reproduce (for bugs)
[example_data.xlsx](https://github.com/crealytics/spark-excel/files/5152348/example_data.xlsx)
Load the attached file data. This is not the real data I'm using but illustrates the problem. Using this code:

    val columns =
      (1 to numberOfColumnsToParse).map(n => StructField(s"$n", StringType))
    val df = spark.read
      .format("com.crealytics.spark.excel")
      .option("dataAddress", s"$da")
      .option("useHeader", header)
      .option("treatEmptyValuesAsNulls", "true")
      .option("inferSchema", "false")
      .option("addColorColumns", "false")
      .option("ignoreLeadingWhiteSpace", "true")
      .option("ignoreTrailingWhiteSpace", "true")
      // .option("maxRowsInMemory", 20)
      .schema(StructType(columns))
      .load(FileUtils.normalizeFilePath(path))

This is result:
![image](https://user-images.githubusercontent.com/24858593/91761609-1a337d00-eba3-11ea-864c-bb5ec4f83fa1.png)

There should be empty rows above metadata. In the actual data there are around 45 such occurrences and it's throwing the dataframe off.

## Your Environment
"com.crealytics" %% "spark-excel" % "0.12.0"
val sparkVersion = "2.4.4"
val hadoopVersion = "3.1.2"

This might just be a spark issue because when a csv file is handled the rdd had to be parsed and add an empty row in like so: 
      .readFileAsStringRDD(path)
      .toDS()
      .map(rowString => if (rowString.isEmpty) "$$$emptyRow" else rowString)

I'll admit that I'm new to both spark and scala so it's possible there is an obvious solution that I'm not seeing and this is a dumb question. I'm also willing to add more information as requested. Any help is appreciated!

nightscape:
This probably wont solve your problem, but could you first try upgrading to the latest version `0.13.1`?

Bryce21:
I downgraded to get around this issue: https://github.com/crealytics/spark-excel/issues/122. I ended up using POI and parsing the file directly if anyone is interested. Which also didn't handle empty rows, columns, and cells well interesting enough. Should I close this?

[...]

PR #210, opened 2020-02-20
WIP: Make Excel a Spark FileFormat

nightscape:

Issue #185, opened 2019-12-19
Cannot read/load folder in spark


yuzhangyi:
Today when I try to load one folder which contains many excel files, it is not supported in current spark_excel, can we support it?

Example: spark.read.format("com.crealytics.spark.excel").option("useHeader", "true").option("inferSchema", "true").load("wasbs://inputdata@adfdataflowtest.blob.core.windows.net/data/excel")

Error: java.io.FileNotFoundException: wasbs://inputdata@adfdataflowtest.blob.core.windows.net/data/excel is a directory not a file.

yuzhangyi:
> Today when I try to load one folder which contains many excel files, it is not supported in current spark_excel, can we support it?
> 
> Example: spark.read.format("com.crealytics.spark.excel").option("useHeader", "true").option("inferSchema", "true").load("wasbs://inputdata@adfdataflowtest.blob.core.windows.net/data/excel")
> 
> Error: java.io.FileNotFoundException: wasbs://inputdata@adfdataflowtest.blob.core.windows.net/data/excel is a directory not a file.

yuzhangyi:
**Scenario**: It has one folder which contains multiple excel files (for example, test01.xlsx, test02.xlsx, test03.xlsx...), now the customer wants to load/read data with sheetName "sheet01" from all up excel files. The spark_excel will try to read data with sheetName "sheet01" from test01.xlsx, test02.xlsx... one by one, if the schema is different compared with first one, the remaining read will fail directly with error "schema is inconsistent among different excel files". If the schema of sheet01 is the same in all up excel files, we merge and return all data to customers.

**Question**: Can I help to add up function in spark_excel if no one does it recently? If yes, can you help share the process that I need to follow? Thanks!

nightscape:
@yuzhangyi I have worked on code to do this:
https://github.com/crealytics/spark-excel/pull/153
Unfortunately, some tests are still failing and I haven't had time to fix everything.
You can try to build the PR locally and see if it works for you.

yuzhangyi:
> @yuzhangyi I have worked on code to do this:
> #153
> Unfortunately, some tests are still failing and I haven't had time to fix everything.
> You can try to build the PR locally and see if it works for you.

Thanks for your information, @nightscape , I will have a try.

wenyangchou:
How can I use it .   Would you give me a demo if you tried succeed

burbeci:
Hey,
[...]

PR #389, opened 2021-06-27
#210 File format v2

quanghgx:
With Spark Data Source API V2, continue the work from #210

Spark-excel V2 intended to be go-to solution when analysis data in excel format, or at least on-boarding data from excel format.
* Access through spark data frame API, so pyspark, scala or any Spark supported languages will have same feature sets
* Reduce external dependencies to Apache POI and its core dependencies

Detail

1. Keep current V1 by adding V2 to a sub folder. Couple of classes are deliberately forked
2. Added support for glob pattern & partitioning folder structure #74 #97 #185 #371
3. Fixed input_file_name #52
4. Added excel row-number column #40 #59 #115
5. [Incompatible with V1] Stricter numerical types (Integer, Long and Double) in schema inferring. Issue #162
6. Added option to keep undefined-row #285 with idea from [upstream guide](http://poi.apache.org/components/spreadsheet/quick-guide.html#Iterator)
7. Writing to multiple files with dataAddress
8. No changing on dependencies, testing dependencies nor build scripts

Points for further improvement

* Support template based in writing: a template file can be specified, and all output files will be based on this file. This also helps for the case of writing to a table by name
* Writing to a single existing file by table-name is not fully implemented (yet)
* User manual for common use cases and explain the meaning and logic of each parameters. Maybe a wiki page or document on a good documentation platform. (prefer nice & clear document)


quanghgx:
Hi @nightscape please help suggest how to activate workflow for this pull request.
Sincerely,

nightscape:
Hey @quanghgx, amazing work you're doing here!!
I've enabled the workflow, the tests have failed due to some imports it couldn't find.
Is your plan to keep the `v2` package or would you replace the old code with `v2`?
I'd have a bias towards the latter (with a series of pre-release builds so that people can test your code) because it makes reviewing easier and we don't have to maintain two code bases.
What do you think?

quanghgx:
Hi @nightscape  

Thanks for enabling the workflow for this PR.

My bad, I only tested with spark 3.1.2. I've just done a quick look around, it is not easy to import different changed-spark-packages with a single code base. In the worst case, we can use a branch for each spark-version.

We may need to keep both, or at least maintain the current version in a dedicated branch. For spark-excel users, we don't want to break their working system when we introduce a new upgrade. At the same time, clearly list the supported spark versions and its expected lifetime.

My 2 cents. Will follow your lead in all of these routes.
Thanks once again, @nightscape

quanghgx:
Hi @nightscape 

For this MR, I can revise it to be compatible with spark 3.0.0+; But we still need to handle the API changes between spark 2.4.* and 3.*. The main issue 

Would like to have a call with you to see how we can go forward with this MR. And how to maintain multiple spark versions in general. I will prepare the list of approaches that I can come up with, mostly from similar open source projects. What do you think about this?

nightscape:
Hey @quanghgx, I appreciate all the work you invest into this project!!
Yes, let's have a call about this 👍 
I contacted you on LinkedIn, we can arrange the details there.

quanghgx:
Hi @nightscape, 
Thanks for reaching out to me on Linkedin. As noted there, would like to have you suggestion, and will summary main points in this PR before implement.
Sincerely,

quanghgx:
rebased to the main branch

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -0,0 +1,405 @@
...
+
+      /* Other files also be utilized (lazily) for field types, reuse field name
+         from the first file */
+      val numberOfRowToIgnore = if (options.header) (options.ignoreAfterHeader + 1) else 0
+      paths.tail.foreach(path => {
+        rows ++= excelHelper.getRows(conf, path).drop(numberOfRowToIgnore)
+      })
+
+      /* Limit numer of rows to be used for schema infering */
+      rows = if (options.excerptSize.isDefined) rows.take(options.excerptSize.get) else rows
```
nightscape, line 295:
```suggestion
      rows = options.excerptSize.foldLeft(rows)(_ take _)
```

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -0,0 +1,405 @@
...
+          rows = Iterator(headerRow) ++ rows
+          r
+        }
+
+      /* Other files also be utilized (lazily) for field types, reuse field name
+         from the first file */
+      val numberOfRowToIgnore = if (options.header) (options.ignoreAfterHeader + 1) else 0
+      paths.tail.foreach(path => {
+        rows ++= excelHelper.getRows(conf, path).drop(numberOfRowToIgnore)
+      })
```
nightscape, line 292:
```suggestion
      rows = paths.tail.foldLeft(rows) { case (rs, path) => 
        rs ++ excelHelper.getRows(conf, path).drop(numberOfRowToIgnore)
      )
```

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -0,0 +1,405 @@
...
+    extends DataWriterFactory[InternalRow] {
+  override def createDataWriter(partitionId: Int, taskId: Long, epochId: Long): DataWriter[InternalRow] =
+    new ExcelDataWriter(options, (new Path(path, f"part-$partitionId%06d-$taskId%09d.xlsx")).toString(), schema)
+}
+
+class ExcelDataWriter(val options: ExcelOptions, val path: String, val schema: StructType)
+    extends DataWriter[InternalRow] {
+
+  private val gen = new ExcelGenerator(path, schema, new Configuration(), options)
+  if (options.header) { gen.writeHeaders() }
```
nightscape, line 393: This means that `ExcelDataWriter` starts writing as soon as it's instantiated, not when the `write` method is called.
Not sure what the contract of a `DataWriter` is, if it can write headers etc. before the first `write` call.

nightscape:
Looks good to me! The only thing I'm not sure about is https://github.com/crealytics/spark-excel/pull/389#discussion_r690484304

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
quanghgx: Hi @nightscape 
Frankly, I haven't thought about this before. Basically, I was just followed same approach with built-in [CsvOutputWriter.scala](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/execution/datasources/csv/CsvOutputWriter.scala#L42)
Thinking about this now, I think, when writing with spark data frame, something like this:
```python
df.write.format("excel").save("/an/example/path")
```
spark is lazy by its natural, above paragraph, including how the df is constructed, will be prepared into an execution plan and all this will be actually materialized in the .save(..) action.
At the time of OutputWriter is constructed (on spark executor node) all details about the writing are ready (with header or not, to which path etc). Even if df is empty, writing a header line still makes sense.

nightscape:
This is good to merge from my side!
@quanghgx let me know if I can push the button 👍

quanghgx:
Thank you so much @nightscape 
Added examples to both wiki and a notebook in docs folder. Please squash commits message when you do the merge. It's ready!

PR #399, opened 2021-08-10
Support error type for formula cell

EnverOsmanov:
When a cell has formula inside LibreOffice shows it empty, but spark-excel silently skips(!) these rows.

@waiyan1612 could you explain, please, why `PlainNumberReadSuite` expects to skip rows? See failed test cases.

### DIFF
--- a/src/main/scala/com/crealytics/spark/excel/DataColumn.scala
+++ b/src/main/scala/com/crealytics/spark/excel/DataColumn.scala
@@ -67,6 +67,7 @@ class HeaderDataColumn(
             case CellType.NUMERIC => Option(cell.getNumericCellValue)
             case CellType.STRING =>
               parseNumber(Option(cell.getRichStringCellValue).map(_.getString))
+            case CellType.ERROR => Some(0.0)
           }
       }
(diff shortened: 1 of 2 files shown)


waiyan1612:
Hi @EnverOsmanov, when I was developing the test suite, I just followed the existing behaviour of the `inferSchema`. That's why in the suite, using `inferSchema=true` is expecting 4 rows but `inferSchema=false` is expecting 5 rows. 

Perhaps @nightscape can provide a better context. We will also need to consider the backward compatibility of this change.

nightscape:
The "correct" solution will land with https://github.com/crealytics/spark-excel/pull/389 where one can specify a [`ParseMode`](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/util/ParseMode.scala) which determines how to handle rows with corrupt data.
I'd be ok with merging this as a temporary solution after the tests are adapted accordingly.

nightscape:
@EnverOsmanov LGTM! Feel free to merge 👍

PR #408, opened 2021-09-01
For Spark DataSource API V2, spark-excel jar depends on spark-version

quanghgx:
Overview: 
 
- By using Spark Data Source API V2, which changed its signature between 2.4.*, 3.0.0 and 3.1.x, made spark-excel jar depends on specific version of spark. This is similar to [spark-testing-base](https://mvnrepository.com/artifact/com.holdenkarau/spark-testing-base)
- Users can still self-assemble the spark-excel jar file for their specific spark environment.
 
Spark Excel, however, also needs to publish jars with spark versions, for users who choose to use pre-assembled jars in public Maven repositories.
 
Changes:
- Prefix spark-excel version with _spark-version_
- Minor update on the Readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,14 +28,14 @@ You can link against this library in your program at the following coordinates:
 ```
 groupId: com.crealytics
 artifactId: spark-excel_2.12
-version: 0.14.0
+version: <spark-version>_0.14.0
 ```
 
 ### Scala 2.11
 ```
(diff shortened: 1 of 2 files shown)


quanghgx:
Hi @nightscape, Please help review this MR.

nightscape:
Looks good to me!

Issue #377, opened 2021-05-19
Update README :) 


aestela-brighthealthplan:
Hello!

I ran into several issues with this library because I was using version 0.13.1 as specified on the README. I finally came across the closed issue that addressed what I was seeing, and the release that fix was pushed with. Only then did I think to go to Maven repo and see that you had later versions. 

Could you update the README to show the latest version, or point to the maven repo to find the latest version? (And any other cleanup). Hopefully that will save others the few hours it cost me. 

Thanks!

nightscape:
Hi @aestela-brighthealthplan,

that would be a nice first PR 😃 
You can check how the badge in README.md is set for the SNAPSHOT version, add a similar badge for the stable version and mention in the README, that the above versions should be used.

AlbertUlysses:
Hi @nightscape 
If no one is working on this, I can help with it :smile:

aestela-brighthealthplan:
So sorry I missed your response @nightscape ! All yours @AlbertUlysses

nightscape:
Hi @AlbertUlysses, depending on your familiarity with SBT and Github Pages, it would be very nice to publish a micro-site using e.g. https://47degrees.github.io/sbt-microsites/
I assume this would allow interpolating the currently published versions into the HTML, so it can't be forgotten anymore 😉

quanghgx:
Resolve this issue, as #420 is merged. Thanks to @pjfanning

Issue #331, opened 2020-12-10
Nexus badge does not update anymore


nightscape:
## Expected Behavior
After pushing a new release to Maven Central, the badge at the top of the README should display the newest version.

## Current Behavior
The badge still shows a rather old version `0.13.3-9...SNAPSHOT`.

## Possible Solution
Research for possible problems with the badge and check if someone else has found a solution.

quanghgx:
Resolved by @pjfanning in https://github.com/crealytics/spark-excel/pull/420

PR #420, opened 2021-09-18
change to maven central badge (nexus one broken)

pjfanning:


quanghgx:
This is great, @pjfanning
Look like this will resolve #377 and #331 . If that is the case, please close these 2 issues after this PR is merged.

pjfanning:
@quanghgx this relates to #331 but #377 has more to it than just the badge - I don't have rights to close either issue

PR #465, opened 2021-11-25
Fix build by upgrading apache.poi to 5.1 and explicitely adding ooxml providers

clemensvonschwerin:
Hey crealytics/spark-excel team,

thank you for upgrading to poi version 5. It resolves our issues with reading large excel files in spark. Unfortunately, in the current version of your jar shadeio.poi seems to be missing. With the upgrade to poi version 5.1 and the small addition to initialization in the code the jar built from sources works like a charm for us. Would be great if we could get this to mainline so everybody profits.

Best,

Clemens

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,8 +21,8 @@ libraryDependencies ++= Seq("org.slf4j" % "slf4j-api" % "1.7.32" % "provided")
   .map(_.excludeAll(ExclusionRule(organization = "stax")))
 
 shadedDeps ++= Seq(
-  "org.apache.poi" ^ "poi" ^ "5.0.0",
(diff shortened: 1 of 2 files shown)


nightscape:
0.15.1 is on its way 👍 
https://github.com/crealytics/spark-excel/releases/tag/v0.15.1

nightscape:
Hey @Edgar-Kalkowski, nice to see you around 😄

EnverOsmanov:
That's fantastic news! I'm glad there is other way to add providers.

With "maxRowsInMemory" it works perfectly for me. But with default reader I still have some issues. Let's see who else got it.

Issue #463, opened 2021-11-24
spark-excel doesn't take into account the 'spark.sql.datetime.java8API.enabled' conf added in spark 3


cristichircu:
Spark 3 added support for working with java.time.LocalDate/Instant instead of java.sql.Date/Timestamp by setting the [spark.sql.datetime.java8API.enabled](https://github.com/apache/spark/blob/3565c3aff9be6cc4e9d6faf9625c0c40ea5fa8d4/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L3014) to true. However this doesnt reflect when reading data from excel using-spark excel and that will cause issues.

## Expected Behavior
Read date and timestampt using the new api when the flag is enabled

## Current Behavior
Dates and timestamps are read using the old API. Exception I ran into is somehting like:
```
java.lang.RuntimeException: Error while encoding: java.lang.RuntimeException: java.sql.Date is not a valid external type for schema of date
.....
if (assertnotnull(input[0, org.apache.spark.sql.Row, true]).isNullAt) null else staticinvoke(class org.apache.spark.sql.catalyst.util.DateTimeUtils$, DateType, localDateToDays, validateexternaltype(getexternalrowfield(assertnotnull(input[0, org.apache.spark.sql.Row, true]), 4, Date), DateType), true, false) AS Date#3531
.....
	at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$Serializer.apply(ExpressionEncoder.scala:213)
	at org.apache.spark.sql.catalyst.encoders.ExpressionEncoder$Serializer.apply(ExpressionEncoder.scala:195)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:461)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:755)
	at org.apache.spark.sql.execution.columnar.DefaultCachedBatchSerializer$$anon$1.hasNext(InMemoryRelation.scala:118)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
	at org.apache.spark.storage.memory.MemoryStore.putIterator(MemoryStore.scala:221)
[...]

PR #475, opened 2021-12-05
Bug fix for #463: Handle Date and Decimal types

quanghgx:
* Handle Date and Decimal types
* Data for unit-test is from @cristichircu in #463


src/main/scala/com/crealytics/spark/v2/excel/ExcelParser.scala
```diff
@@ -247,7 +254,7 @@ class ExcelParser(dataSchema: StructType, requiredSchema: StructType, val option
               datum.getCellType match {
                 case CellType.NUMERIC | CellType.FORMULA => {
                   /* Excel to spark precision */
-                  datum.getNumericCellValue * 1000
```
nightscape, line 250: Why is this not needed anymore? Does Spark do something differently if it gets a Long value?

quanghgx: Hi Martin,

It was incorrect before. This code section gets called when a cell is not formatted as Date or Time, however, users pass a custom schema and want to load that cell value as Timestamp, so there is no binding logic to multiply it to 1000.

We haven't seen a case like this in real life yet. And only when handling DateType, I recognized this issue.

nightscape:
LGTM! Merge at your convenience 👍

cristichircu:
@quanghgx are you waiting for something from my side or are you busy with somehting else? Just trying to make sure I didn't miss a request or something :) Thank you!

nightscape:
@quanghgx @cristichircu I just went ahead and merged this. It's got tests so we should be good 👍

quanghgx:
Thank you Martin.
Hi @cristichircu , it's just that I am lacking the time to put it forward.

PR #491, opened 2021-12-21
480

cristichircu:
thought this was a good place to add the providers since it's used by all the spark versions plus this should be a singleton so there's no danger of calling it multiple times and adding the providers multiple times
thought of going for the bonus and unifying the poi initialization here with v1 but I couldn’t figure out where common code between the 2 should be. Any suggestions are welcomed :)

### DIFF
--- a/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
+++ b/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
@@ -11,23 +11,18 @@
   */
 package com.crealytics.spark.v2.excel
 
-import org.apache.hadoop.conf.Configuration
-import org.apache.hadoop.fs.FileSystem
-import org.apache.hadoop.fs.Path
-import org.apache.poi.ss.usermodel.Cell
-import org.apache.poi.ss.usermodel.CellType
-import org.apache.poi.ss.usermodel.DataFormatter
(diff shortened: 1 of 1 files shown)


nightscape:
Afaik, the body of `object`s is evaluated only when a method on the object is called the first time.
It would be nice if we could be more explicit about this, e.g. by putting the registration inside a method that gets called at an appropriate time.
The question is if registering the factories is idempotent and we can call this method as many times as we want, or if that would actually add the factories to some list multiple times.
@cristichircu could you check that out or see if there's a way to not register providers that have already been registered?

cristichircu:
Sorry, I was out for the holidays and I wasn't able to look into this before today.

As far as I can tell poi holds providers in an ArrayList so adding them isn't idempotent. 
The good news is that it returns when the first provider accepts the file so it wouldn't matter if you have a provider registered multiple times. However, this is an implementation detail that could change at any time and I wouldn't base my solution on it.
You could also try to remove a provider before adding it. You don't currently have a way of checking if a provider exists, but the current implementation of remove doesn't throw an exception if it doesn't. Again... implementation detail that I'm reluctant to count on.

Any thoughts?

nightscape:
Imo it would be best to `removeProvider` immediately before `addProvider`. That way we make sure that the required `WorkbookFactory`s are in the list exactly once.

Issue #480, opened 2021-12-08
Reading a table from an excel file with format("excel") throws IOException in the 0.15.1 version


nicoricimadalina:
Previously working code for reading an excel file with the v2 format is failing after upgrading the library from 0.14.0 to the 0.15.1 version. 

## Expected Behavior
Reading with the format("excel") still works as with previous versions.

## Current Behavior
Reading my excel file works for the format("com.crealytics.spark.excel") but with the format("excel") fails with the exception:

```
java.io.IOException: Your InputStream was neither an OLE2 stream, nor an OOXML stream or you haven't provide the poi-ooxml*.jar in the classpath/modulepath - FileMagic: OOXML, having providers: []
	at shadeio.poi.ss.usermodel.WorkbookFactory.wp(WorkbookFactory.java:309)
	at shadeio.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:208)
	at shadeio.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:172)
	at com.crealytics.spark.v2.excel.ExcelHelper.getWorkbook(ExcelHelper.scala:107)
	at com.crealytics.spark.v2.excel.ExcelHelper.getRows(ExcelHelper.scala:122)
	at com.crealytics.spark.v2.excel.ExcelPartitionReaderFactory.readFile(ExcelPartitionReaderFactory.scala:74)
	at com.crealytics.spark.v2.excel.ExcelPartitionReaderFactory.buildReader(ExcelPartitionReaderFactory.scala:61)
	at org.apache.spark.sql.execution.datasources.v2.FilePartitionReaderFactory.$anonfun$createReader$1(FilePartitionReaderFactory.scala:29)
	at scala.collection.Iterator$$anon$10.next(Iterator.scala:461)
	at org.apache.spark.sql.execution.datasources.v2.FilePartitionReader.getNextReader(FilePartitionReader.scala:106)
	at org.apache.spark.sql.execution.datasources.v2.FilePartitionReader.next(FilePartitionReader.scala:42)
	at org.apache.spark.sql.execution.datasources.v2.PartitionIterator.hasNext(DataSourceRDD.scala:79)
	at org.apache.spark.sql.execution.datasources.v2.MetricsIterator.hasNext(DataSourceRDD.scala:112)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
[...]

PR #513, opened 2022-01-11
fix for https://github.com/crealytics/spark-excel/issues/480

cristichircu:
create a new pull request with suggestions from https://github.com/crealytics/spark-excel/pull/491

### DIFF
--- a/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
+++ b/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
@@ -203,10 +203,12 @@ object ExcelHelper {
   def apply(options: ExcelOptions): ExcelHelper = new ExcelHelper(options)
 
   def configureProviders() : Unit = {
-    WorkbookFactory.removeProvider(classOf[HSSFWorkbookFactory])
-    WorkbookFactory.addProvider(new HSSFWorkbookFactory)
+    synchronized {
+      WorkbookFactory.removeProvider(classOf[HSSFWorkbookFactory])
+      WorkbookFactory.addProvider(new HSSFWorkbookFactory)
 
-    WorkbookFactory.removeProvider(classOf[XSSFWorkbookFactory])
-    WorkbookFactory.addProvider(new XSSFWorkbookFactory)
+      WorkbookFactory.removeProvider(classOf[XSSFWorkbookFactory])
+      WorkbookFactory.addProvider(new XSSFWorkbookFactory)
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


cristichircu:
@nightscape Is this what you had in mind?
I was thinking about making a similar change for v1 and adding a configureProviders method and calling it before [workbook reader initialization](https://github.com/crealytics/spark-excel/blob/07aea8290e9cd9797b7320fb44121929ca225021/src/main/scala/com/crealytics/spark/excel/DefaultSource.scala#L22) or in [withWorkbook](https://github.com/crealytics/spark-excel/blob/07aea8290e9cd9797b7320fb44121929ca225021/src/main/scala/com/crealytics/spark/excel/WorkbookReader.scala#L15). Thoughts?

Thnks!

nightscape:
That's exactly what I had in mind 👍 
I would say the workbook reader initialisation would be a good place.

cristichircu:
looks like we're running into some ConcurrentModificationExceptions caused by removeIf. Do you think it’s worth synchronizing this? I would revert to the initial version (like it is for v1)

nightscape:
Can you maybe try putting it at the [start of `createRelation`](https://github.com/crealytics/spark-excel/blob/07aea8290e9cd9797b7320fb44121929ca225021/src/main/scala/com/crealytics/spark/excel/DefaultSource.scala#L22)?
Afaik, this method only gets called when a `spark.read.format("excel")` is being run.
The `synchronized` might still make sense though if `read` is called in parallel.

cristichircu:
The ConcurrentModificationException was for v2. If I understand correctly your suggestion is only applicable for v1.
Should I understand that v2 is ok the way it is and your suggestion is related to [my comment](https://github.com/crealytics/spark-excel/pull/513#issuecomment-1009785825) earlier regarding the proposed changed for v1?

nightscape:
Ah sorry, confusion from my side. I thought there is a `createRelation` in v2 as well.
I'm not yet familiar with the v2 implementation as this was done by @quanghgx.
@quanghgx would you know a good place where the registration of `WorkbookFactory`s could be done in v2?
Best in a method that does not get called too often / concurrently.

cristichircu:
@quanghgx can you please share your thougths on this when you get a chance? Thank you!

quanghgx:
Hi @nightscape and @cristichircu 
Sorry for missing this thread. Let me double check and get back to you two this weekend.
Sincerely,

cristichircu:
Hi @quanghgx ! Did you get a chance to have a look at this? Thanks!

quanghgx:
Hi @nightscape and @cristichircu 
1+ with v2/excel/ExcelHelper.scala, this seems the most practical position to put this initialization.
Sincerely,

Issue #550, opened 2022-02-17
LocalDate and Instant aren't handled correctly on write in v2


cristichircu:
Spark 3 added support for working with java.time.LocalDate/Instant instead of java.sql.Date/Timestamp by setting the [spark.sql.datetime.java8API.enabled](https://github.com/apache/spark/blob/3565c3aff9be6cc4e9d6faf9625c0c40ea5fa8d4/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L3014) to true. However this doesn't reflect when writing data to excel using-spark excel v2 and that will cause issues.

## Expected Behavior
Write the correct date and timestamp values to excel

## Current Behavior
Incorrect values are written to excel and they are displayed as 1/1/1900 or whatever the date format is

## Possible Solution
Looks to me like support for the new datatypes needs to be added in [ExcelGenerator](https://github.com/crealytics/spark-excel/blame/bce698336632b8ea3579522aca4b8ff31dfe67a5/src/main/scala/com/crealytics/spark/v2/excel/ExcelGenerator.scala#L107-L117)

## Steps to Reproduce (for bugs)
Write dates and timestamps havind the spark.sql.datetime.java8API.enabled flag set to true

## Your Environment
> Include as many relevant details about the environment you experienced the bug in
* Spark version and language: java 11 and spark 3.1.1/3.2.1
* Spark-Excel version: com.crealytics:spark-excel_2.12:3.1.2_0.16.4

cristichircu:
Another thing I noticed in ExcelGenerator and that I mentioned [here](https://github.com/crealytics/spark-excel/issues/126#issuecomment-1042933375) is that decimals are always formatted with scientific notation. While that isn't an issue if the consumer is another system, it's not ideal if a user wants to have a look. Could we set the decimal format to General? We could also use the [usePlainNumberFormat](https://github.com/crealytics/spark-excel/blob/bce698336632b8ea3579522aca4b8ff31dfe67a5/src/main/scala/com/crealytics/spark/v2/excel/ExcelOptions.scala#L102) flag so users can turn this behavior on.

Thank you!

cristichircu:
issue should be fixed in https://github.com/crealytics/spark-excel/releases/tag/v0.17.0

PR #552, opened 2022-02-18
fix for https://github.com/crealytics/spark-excel/issues/550

cristichircu:
Hey @nightscape, @quanghgx!
I tried a fix for 550 but I can't seem to get it working for timestamps. The current code doesn't seem to generate a valid timestamp when writing excel files for v2 (as you can see from the test). I also tried `DateTimeUtils.toJavaTimestamp(row.getLong(ordinal))` which seems like it should be the clean solution and a bunch of variations on it but without success. Do you guys have any idea how I could get this working?

Thanks!

### DIFF
--- a/src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
+++ b/src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
@@ -20,6 +20,7 @@ import java.util
 import scala.collection.JavaConverters._
 import java.nio.file.Files
 import java.sql.{Date, Timestamp}
+import java.time.{Instant, LocalDate}
 
 import com.holdenkarau.spark.testing.DataFrameSuiteBase
 
@@ -80,12 +81,20 @@ object WriteAndReadSuite {
     )
   )
 
-  val expectedData_03: util.List[Row] = List(
+  val expectedData_03_sql: util.List[Row] = List(
     Row(1, Date.valueOf("2021-10-01"), Timestamp.valueOf("2021-10-01 01:23:45")),
     Row(2, Date.valueOf("2021-11-01"), Timestamp.valueOf("2021-11-01 11:23:45")),
     Row(3, Date.valueOf("2021-10-11"), Timestamp.valueOf("2021-10-11 01:23:45")),
     Row(4, Date.valueOf("2021-11-11"), Timestamp.valueOf("2021-11-11 01:23:05")),
     Row(5, Date.valueOf("2022-10-01"), Timestamp.valueOf("2022-10-01 16:23:45"))
+  ).asJava
+
+  val expectedData_03_time: util.List[Row] = List(
+    Row(1, LocalDate.parse("2021-10-01"), Instant.parse("2021-10-01T01:23:45Z")),
+    Row(2, LocalDate.parse("2021-11-01"), Instant.parse("2021-11-01T11:23:45Z")),
+    Row(3, LocalDate.parse("2021-10-11"), Instant.parse("2021-10-11T01:23:45Z")),
+    Row(4, LocalDate.parse("2021-11-11"), Instant.parse("2021-11-11T01:23:05Z")),
+    Row(5, LocalDate.parse("2022-10-01"), Instant.parse("2022-10-01T16:23:45Z"))
   ).asJava
 }
 
@@ -149,7 +158,7 @@ class WriteAndReadSuite extends AnyFunSuite with DataFrameSuiteBase with ExcelTe
   test("write then read java.sql.date and java.sql.timestamp") {
     val path = Files.createTempDirectory("spark_excel_wr_03_").toString()
     spark.conf.set(DATETIME_JAVA8API_ENABLED, false)
-    val df_source = spark.createDataFrame(expectedData_03, userDefinedSchema_03).sort("Id")
+    val df_source = spark.createDataFrame(expectedData_03_sql, userDefinedSchema_03).sort("Id")
     df_source.write.format("excel").mode(SaveMode.Append).save(path)
 
     val df_read = spark.read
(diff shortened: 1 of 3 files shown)


quanghgx:
Hi @cristichircu 
Thank you so much for handling these two issues.

I've just rerun the CI build, it seems that, original failed build is unrelated to your fix (that was something about loading sbt plugin, temporary I think).

With the new build, there are two failed test-cases. I cloned your repos (main branch, https://github.com/cristichircu/spark-excel) and got the same error as in github CI/CD.
Screenshot from 2022-02-20 20-52-10

Please help take another look?

cristichircu:
Hi @quanghgx! As I mentioned in the comment above, I wasn't able to fix write timestamp (but I added tests to point out that the current implementation isn't working). That's why the 2 tests are failing. Could you please point me in the right direction? I tried multiple variations starting from `DateTimeUtils.toJavaTimestamp(row.getLong(ordinal))` but nothing seemed to work. Any help would be highly appreciated.

Thanks!

quanghgx:
Got it, sorry for my misunderstanding. Will get back to you soon.

cristichircu:
Hi, @quanghgx ! Did you get a chance to have a look at this? Thanks!

cristichircu:
Hello @nightscape , @quanghgx!
Today I finally managed to have another look at this and found the timestamp issue to be caused by the default format it had. The current format contained some illegal characters and that was causing the issues. I fixed that by setting another default format (the default timestamp pattern from spark catalyst).

Currently some tests using LocalDate and Instant are failing for some spark versions because of how spark switches between Date, Timestamp and LocalDate, Instant internally. 
Is there a way to run specific tests for specific spark versions? Should I just comment out the test and let users run it manually when required? Should I delete it? Looking forward for your suggestions.

Thanks,
   Cristi

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
```diff
@@ -126,4 +155,42 @@ class WriteAndReadSuite extends AnyFunSuite with DataFrameSuiteBase with ExcelTe
...
+    val df_read = spark.read
+      .format("excel")
+      .schema(userDefinedSchema_03)
+      .load(path)
+      .sort("Id")
+
+    assertDataFrameEquals(df_source, df_read)
+
+    /* Cleanup, should after the checking */
+    spark.conf.unset(DATETIME_JAVA8API_ENABLED)
```
nightscape, line 173: Can we `.get` the value before and then `.set` to that? With the current implementation, we would be losing the previous value.

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
```diff
@@ -67,6 +72,30 @@ object WriteAndReadSuite {
...
+  val userDefinedSchema_03 = StructType(
+    List(
+      StructField("Id", IntegerType, nullable = true),
+      StructField("Date", DateType, nullable = true),
+      StructField("Timestamp", TimestampType, nullable = true)
+    )
+  )
+
+  val expectedData_03_sql: util.List[Row] = List(
+    Row(1, Date.valueOf("2021-10-01"), Timestamp.valueOf("2021-10-01 01:23:45")),
```
nightscape, line 85: You could store the data as `Tuple3[Int, String, String]` once and then convert the two `String` entries into the required datatypes in the corresponding tests. Also, no need to convert to a Java `util.List[Row]`. Spark has a [`.createDataFrame` version that takes a `Seq[A <: Product]`](https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/SparkSession.html#createDataFrame[A%3C:Product](data:Seq[A])(implicitevidence$3:reflect.runtime.universe.TypeTag[A]):org.apache.spark.sql.DataFrame) for which should work with a `List[Tuple3[...]]`.

nightscape:
Hey @cristichircu, I made a few minor comments. Not sure if the `spark.conf.unset(DATETIME_JAVA8API_ENABLED)` accidentally breaks the tests by removing the previously set value?

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
cristichircu: done

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
cristichircu: partially done. I wasn't able to get it working nicely with .createataFrame(Seq[A]) since that doesn't take a Struct and removing the Struct caused some other issues (after repplacing "Id" with "_1" as the column name)

cristichircu:
Thanks for your feedback @nightscape!
That test still fails for spark versions before 3.0 when that flag and functionality was added. I tried figuring out if we're in an earlier version based on the config having a default value but it looks like that didn't work. Any pointers on how I could handle a test that should only run on versions of spark after 3.0?

cristichircu:
Ended up just canceling the test if spark.version.startsWith("2."). Everything seems to be fine now. Please let me know if there's anything else I can do to get this approved

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
```diff
@@ -48,25 +54,19 @@ object WriteAndReadSuite {
...
-      StructField("Day", LongType, true),
-      StructField("Month", LongType, true),
-      StructField("Customer ID", StringType, true),
-      StructField("Customer Name", StringType, true),
-      StructField("Standard Package", IntegerType, true),
-      StructField("Extra Option 1", IntegerType, true),
-      StructField("Extra Option 2", IntegerType, true),
-      StructField("Extra Option 3", LongType, true),
-      StructField("Staff", StringType, true)
+      StructField("Id", IntegerType, nullable = true),
```
quanghgx, line 57: For bug fix, if it is possible, please add a test-case instead of modify the existing one?
If the existing one is incorrect, we may need to add another MR dedicated for that.
In long term, this is just to avoid a regression issue for existing users.

quanghgx (approved):

src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
cristichircu: That's what I did initially.  Afterwards, I noticed that userDefinedSchema_02 and expectedData_02 weren't used and I decided to cleanup. I can revert that if you would like me to.

quanghgx:
1+ with merging, @cristichircu and @nightscape

nightscape:
Done!

cristichircu:
thank you both!

nightscape:
Thank you!

Issue #547, opened 2022-02-15
Writing Excel with partitionBy: requirement failed: The provided partitioning does not match of the table.


christianknoepfle:
## Expected Behavior
When I do a 
```
        dfCsv.write
          .partitionBy("col1")
          .format("excel")
          .option("path", targetDir)
          .option("header", true)
          .mode(writeMode)
          .save()
```
I expect to get files that are located in different subdirectories (similar to csv or parquet).
 
## Current Behavior
The mentioned call results in the following exception and no files are written
```
requirement failed: The provided partitioning does not match of the table.
 - provided: col1
 - table: 
java.lang.IllegalArgumentException: requirement failed: The provided partitioning does not match of the table.
 - provided: col1
 - table: 
	at scala.Predef$.require(Predef.scala:281)
	at org.apache.spark.sql.DataFrameWriter.checkPartitioningMatchesV2Table(DataFrameWriter.scala:748)
	at org.apache.spark.sql.DataFrameWriter.save(DataFrameWriter.scala:350)

```
The exception is not related to the SaveMode (fails with Append and Overwrite)

## Steps to Reproduce (for bugs)

The test code is here:

     https://github.com/christianknoepfle/spark-excel-test

Just clone/download and run the scalatest

## Your Environment
Spark 3.0.3, spark excel 0.16.4

christianknoepfle:
fixed via #562

Issue #554, opened 2022-02-28
dataframe.write.format("excel") supports SaveMode.Overwrite/Append and partitioning 


christianknoepfle:
## Expected Behavior
dataframe.write.format("excel") supports SaveMode.Overwrite/Append and partitioning in the same way as other data sources like csv

## Current Behavior
SaveMode and partitioning doesn't work as expected, see issue #539 and issue #547 

## Possible Solution
Despite the comment in ExcelDataSource I suggest to change the definition of ExcelDataSource 
from
```
class ExcelDataSource extends TableProvider with DataSourceRegister
```
to
```
class ExcelDataSource extends FileDataSourceV2
```
This allows us to remove most of the copied code from spark. 

In addition we provide a simple implementation for the fallbackFileFormat that just supports writing. So we have 

```
class ExcelFileFormat extends FileFormat with DataSourceRegister {
...
  override def prepareWrite ...
...
```
the prepareWrite() is basically taken from ExcelWriteBuilder()

You can see the changes here:: https://github.com/christianknoepfle/spark-excel/pull/1/files

There are some very basic unit tests for savemode and partitioning and they work as expected

I only tested it on spark 3.0.3, so I guess some more work needs to be done. The unit test are mostly ok, I only get some issues on probably non related stuff: PlainNumberReadSuite , ErrorAsStringsReadSuite (I am using Windows for testing, not sure if this is an explanation)  

Please let me know if this makes sense to you and how we can move forward.

Thanks

christianknoepfle:
SHort update. I ran the changes with the ci.yml pipeline and the new tests were ok for >=3.0.3 (the existing tests run all good)

Surprisingely the spark API changed from 3.0.0 to 3.0.3 so the changes did not compile for 3.0.0 (in fact it changed from 3.0.0 to 3.0.1).  I would need to change the structure of the src folders to make it work. To me this does not really make a lot of sense. Is it a prerequisite for the PR to still support 3.0.0 (while 3.0.1/3.0.3 work fine)?

For the 2.x line the new tests failed because I haven't ported the fixes over to it. Is that something the PR would need to include or can we also scope some tests to be 3.x exclusive. Just want to know how much effort is needed.

nightscape:
@christianknoepfle thanks for the effort!
Afair @quanghgx had mentioned the same issues regarding changes from Spark 3.0.0 to 3.0.x.
From my side it's fine to exclude 3.0.0 and target only 3.0.1 upwards.
[...]

Issue #539, opened 2022-02-07
Writing to Excel using v2/excel format: Table does not support overwrite by expression


christianknoepfle:
## Description
When writing a dataframe to excel in overwrite mode I would expect that a file will be created in the given path. 

Following code snippet 

        dfCsv.write
               .format("excel")
                .option("path", targetDir)
                .option("header", true)
                .mode(SaveMode.Overwrite)
                .save()

fails with the exception

```
[info]   org.apache.spark.SparkException: Table does not support overwrite by expression: ExcelTable(excel file:/C:/DATA/Dev/git/spark-excel-test/tmp/v2,org.apache.spark.sql.SparkSession@305a40b2,org.apache.spark.sql.util.CaseInsensitiveStringMap@b7029722,List(tmp\v2),None)  
[info]   at org.apache.spark.sql.execution.datasources.v2.OverwriteByExpressionExec.run(WriteToDataSourceV2Exec.scala:292)
...
```

Doing the same with the V1 API (and the path option points to a dedicated file), it works as expected

    dfCsv.write
        .format("com.crealytics.spark.excel")
        .option("path", targetFile)
        .option("header", true)
        .mode(SaveMode.Overwrite)
        .save()

## Steps to Reproduce (for bugs)
I published a repo with the most basic setup (just spark, spark-excel and scalatest) and a few tests that demonstrate the issue

    https://github.com/christianknoepfle/spark-excel-test

Just pull it and then run sbt test to reproduce the issue (it contains simple read/write test for v1/v2 api)

## Context
We want the V2 API because of two reason:
- *path-option* just points to a path and not an individual file. So same behavior as the other file formats bundled with spark
- need *partitionBy* functionality

## Your Environment
We use Spark 3.0.3, Scala, 0.16.0 spark-excel, Windows (but also Linux on AWS EMR)

### Full trace:
```
[info]   org.apache.spark.SparkException: Table does not support overwrite by expression: ExcelTable(excel file:/C:/DATA/Dev/git/spark-excel-test/tmp/v2,org.apache.spark.sql.SparkSession@305a40b2,org.apache.spark.sql.util.CaseInsensitiveStringMap@b7029722,List(tmp\v2),None)  
[info]   at org.apache.spark.sql.execution.datasources.v2.OverwriteByExpressionExec.run(WriteToDataSourceV2Exec.scala:292)
[...]

PR #562, opened 2022-03-06
V2 write issue (implements proposal in issue #554)

christianknoepfle:
This PR implements the proposal in issue #554 and resolves issue #547 and issue #539. See the proposal for technical approach.
In addition the README.md was enhanced to reflect the changes


nightscape:
Hey @christianknoepfle, thanks for the PR!
I just triggered the Github build.
I'm not too familiar with the v2 implementation, so @quanghgx would have to chime in with a code review here.

christianknoepfle:
The test failures on the spark 3.x are related to reading a partitioned excel file structure and that leads to a crash deep in poi. Unsure how to force this behavior on every test run so it is easier to debug.  Any ideas?

christianknoepfle:
Update: Added a test to get the java.util.ConcurrentModificationException when executing locally by reading a folder with100+ partitions. So I can investigate further.

christianknoepfle:
I found the issue of the java.util.ConcurrentModificationException and added a fix for it.

See here : src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala

The comment block for ExcelHelper.getWorkbook() . The issue was present prior the changes of this PR but didn't show up until now (it occurs rarely and random, but with more partition testcases chances increased). 

Note that there is also an alternative implementation mentioned in the comment block. Since I am not a poi expert, I have no idea if this alternative implementation would be sufficient (I didn't get why the providers are re-registered anyhow, because those factory classes do not have any state)

Could you pls run the workflow so we see if the issue is gone in your pipeline too (it worked on my end, but last time that was the case too)

src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
```diff
@@ -200,15 +200,43 @@ class ExcelHelper(options: ExcelOptions) {
...
+        Singleton.INSTANCE.provider at the same time (i.e. while it is changing).
+
+        This can lead to a java.util.ConcurrentModificationException exception while looping
+        over the mentioned data structure  (code fragment
+        at org.apache.poi.ss.usermodel.WorkbookFactory.wp(WorkbookFactory.java:327)
+
+        In case we need to exchange the provider on each call, we need to move the
+        synchronized so it guards the call to the .wp() function too.
+     */
+    if (!configurationNeeded.getAndSet(true)) {
```
nightscape, line 257: Nice!

christianknoepfle:
after discussing registerProviders() with @cristichircu here (https://github.com/crealytics/spark-excel/issues/480#issuecomment-1065868049) I enabled the code to do registration only once (also moved the call to apply() which seemed to me a better place).  The CI is all green on my end. How can we move forward with this PR?

src/test/scala/com/crealytics/spark/v2/excel/LocalFileTestingUtilities.scala
```diff
@@ -55,11 +54,10 @@ trait LocalFileTestingUtilities {
 
     fixture
   }
-
-  def createTemporaryDirectory(name: String): File = {
+  def createTemporaryDirectory(prefix: String): File = {
```
nightscape, line 57: Could we use the built-in mechanism for creating temp directories?
https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createTempDirectory%28java.nio.file.Path,%20java.lang.String,%20java.nio.file.attribute.FileAttribute...%29

christianknoepfle: done

src/test/scala/com/crealytics/spark/v2/excel/DataFrameWriterApiComplianceSuite.scala
```diff
@@ -0,0 +1,129 @@
...
+
+        dfCsv.write
+          .format("excel")
+          .option("path", targetDir)
+          .option("header", value = true)
+          .mode(writeMode)
+          .save()
+
+        val listOfFiles = getListOfFilesFilteredByExtension(targetDir, "xlsx")
+        assert(listOfFiles.nonEmpty, s"Expected a single file but got more: $listOfFiles")
```
nightscape, line 64: The assertion logic and message don't seem to align.

nightscape:
@quanghgx would you have time to do a review here?

Afaiu it not possible to write a single `.xlsx` file with a specified name, but instead it will always be the default Spark folder structure, right?
So users who want a single file need to do a `.repartition(1)` to get a single file and then move it out of the subdirectory and rename it, or use the v1 writer.
I would be ok with this, as long as users still can write to and read from a single file (which is what I mostly use it for).
In the mid-term, it would be nice to provide some helper function that allows users to read/write single files with the v2 API, so that we can get rid of v1.

cristichircu:
Hi @nightscape! I started a discussion about this sort of functionality a while back: https://github.com/crealytics/spark-excel/discussions/549. Do you think this would be useful to others? Can you please share your thoughts on the proposed approach?

quanghgx:
Hi Martin and @christianknoepfle 
 
Sorry for the long delay from my side. I was a bit tied up in the last couple of months.
 
About single output file (@nightscape): For spark data source 3.x API, if we reuse the same base classes with built in data sources (like CSV, JSON, Parquet e.g), it is not easy to write to a single excel file. However, in concept, these base classes are not the only route. I did have a couple of other spark data sources (for work), those do not depend on spark built-in base classes, so we can write to any structure we need.
 
Supporting multiple spark versions, while reuse spark built-in supportive classes, is quite a burden. To be fair, these built-in classes seem not intended to be reused by external data sources like spark-excel. I do think that we may need to move away from built-in supportive classes and also reduce the number of supported spark versions. Spark-Excel feature set seems stable by now, so if users need spark-excel with older spark-version, they may need to pick an appropriate (older) spark-excel version.
So, @christianknoepfle, 1+ from me if you need to cleanup some fragmented folders.

Sincerely,

build.sbt
```diff
@@ -6,7 +6,7 @@ crossScalaVersions := Seq("2.12.15", "2.13.8")
 
 scalaVersion := crossScalaVersions.value.head
 
-lazy val sparkVersion = "3.2.1"
+lazy val sparkVersion = "3.0.1"
```
quanghgx, line 9: Maybe, keep the latest, stable spark version as the default?

quanghgx:
Given that remain comments are minor.
1+ with merging, @christianknoepfle and @nightscape

nightscape:
Ok, squash-merging it now 👍

Issue #429, opened 2021-10-08
GC overhead while ingesting big excel files


amazoyer:
Hi,

We are having the following stacktrace when ingesting big excel files (over 10000 lines):
```
java.lang.OutOfMemoryError: GC overhead limit exceeded
                at java.util.TreeMap.put(TreeMap.java:577)
                at shadeio.poi.xssf.usermodel.XSSFRow.<init>(XSSFRow.java:80)
                at shadeio.poi.xssf.usermodel.XSSFSheet.initRows(XSSFSheet.java:268)
                at shadeio.poi.xssf.usermodel.XSSFSheet.read(XSSFSheet.java:231)
                at shadeio.poi.xssf.usermodel.XSSFSheet.onDocumentRead(XSSFSheet.java:218)
                at shadeio.poi.xssf.usermodel.XSSFWorkbook.parseSheet(XSSFWorkbook.java:454)
                at shadeio.poi.xssf.usermodel.XSSFWorkbook.onDocumentRead(XSSFWorkbook.java:419)
                at shadeio.poi.ooxml.POIXMLDocument.load(POIXMLDocument.java:184)
                at shadeio.poi.xssf.usermodel.XSSFWorkbook.<init>(XSSFWorkbook.java:288)
                at shadeio.poi.xssf.usermodel.XSSFWorkbookFactory.createWorkbook(XSSFWorkbookFactory.java:97)
                at shadeio.poi.xssf.usermodel.XSSFWorkbookFactory.createWorkbook(XSSFWorkbookFactory.java:147)
                at shadeio.poi.xssf.usermodel.XSSFWorkbookFactory$$Lambda$36/1626927534.apply(Unknown Source)
                at shadeio.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:256)
                at shadeio.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:221)
                at com.crealytics.spark.excel.DefaultWorkbookReader$$anonfun$openWorkbook$1.apply(WorkbookReader.scala:49)
                at com.crealytics.spark.excel.DefaultWorkbookReader$$anonfun$openWorkbook$1.apply(WorkbookReader.scala:49)
                at scala.Option.fold(Option.scala:158)
                at com.crealytics.spark.excel.DefaultWorkbookReader.openWorkbook(WorkbookReader.scala:49)
[...]

PR #567, opened 2022-03-23
Fixes for bug when Spark reads all rows even if it is not required (#429).

kvirund:


kvirund:
Fixes issue #429.

nightscape:
Afair, doing a `.toSeq` on an `Iterator` returns a `Stream` (called `LazyList` starting with Scala 2.13) which is a lazy data structure. If that `Stream` gets traversed only once and the head of the `Stream` is not stored anywhere, the GC can collect all elements that have been processed.
If my explanation is correct, this is somewhat a luck-shot, but you mentioned in the corresponding issue that it helped processing a 1GB file, so good to merge from my side 👍

nightscape:
I only noticed after merging that your PR targets the obsolete `master` branch, but I've applied the same change to the new `main` branch.

nightscape:
@kvirund sorry, I had to take the change out again, it broke a few tests.
This was covered up by the old `master` branch not building properly anymore.
Could you create a new PR against the `main` branch?
The tests will probably fail, but this time for the right reason, which you would need to investigate.

kvirund:
Oh, sorry. I was making and testing my changes in a private repository and there I was using the `main` branch as a base. Changes on Github I implemented in place using the default one. Will redo it soon.

Yes, my understanding aligns with your explanation but I am not an expert in Scala.

kvirund:
Is there any way to run tests before merging PR?

kvirund:
Something is wrong with the solution. Tests are failing. Need to figure out why before creating another PR.

kvirund:
The problem is with this part:https://github.com/crealytics/spark-excel/blob/main/src/main/scala/com/crealytics/spark/excel/WorkbookReader.scala#L16-L18

We take a lazy sequence of rows and close the workbook immediately. StreamingWorkbook has weird behavior: it doesn't throw an error if the workbook is closed. Instead, it reads whatever remains in the cache.

I am not a Spark pro, and wondering if there is some mechanism we can use to understand that rows sequence is not required anymore.

kvirund:
Basically happens something like that:

```java
        try (final InputStream inputStream1 = new FileInputStream("excel-files/very-simple.xlsx"); // has 6 rows on a single sheet
             final Workbook workbook1 = StreamingReader.builder().rowCacheSize(1).bufferSize(4096).open(inputStream1)) {
            workbook1.getSheetAt(0).rowIterator().hasNext();  // populate cache, without it, this snippet would report 0 rows
            workbook1.close();
            int counter = 0;
            for (Row row: workbook1.getSheetAt(0)) {
                ++counter;
            }
            logger.info("Read #{} rows", counter);  // will report 1 row
        } catch (IOException e) {
            e.printStackTrace();
        }
```

nightscape:
Ok, then spark.parallelize seems to be lazy/async. Otherwise it would eagerly evaluate the given Seq before the end of the withWorkbook block is reached.
The improvements you saw for the 1GB file then were probably also caused by just not reading all data, right?
We could slightly improve the situation without breaking anything by doing a toArray instead of a .to[Vector]

kvirund:
> The improvements you saw for the 1GB file then were probably also caused by just not reading all data, right?

Not really. I didn't notice the problem because I was using a bigger cache (200 rows) and for the test, I was reading just the first few rows.

So, improvement was real. And if I don't close the workbook inside the workbook - it would work as expected. But in that case, we may get a resource leak (not closing `AutoClosable`s probably is not the best idea). The problem is I don't know how to understand when RDD finished reading from the lazy `Seq` we construct. Without that, we are obliged to read all rows before closing the workbook which basically means that we have to load into memory the entire sheet. :-(

kvirund:
Probably we can use `finalize` method somehow.

kvirund:
I mean if the `StreamingWorkbook` had `finalize` method - it could invoke `close` when it is not needed anymore. That is how `ZipFile` used in that workbook works.

But that would be a prereq change to a separate repository: to provide a workbook able to finalize itself.

kvirund:
Or, we can add `finalize` method to a custom `Seq` implementation.

Issue #587, opened 2022-04-29
Formulas with errors aren't handled correctly even if the setErrorCellsToFallbackValues/useNullForErrorCells flags are set


cristichircu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When reading a file containing string formulas, if an error is present, the formula is returned instead of null or "", like for String cells, for instance. This is true for v1 and v2. The setErrorCellsToFallbackValues/useNullForErrorCells flags don't help because cell.getCellType first checks if the cell is a formula and, if true, returns FORMULA so it never gets to actually check if it's an error cell.

### Expected Behavior

I'm guessing that you never want to return the formula (for v1 or v2). 
For v1, for consistency, I assume you would want to return null if setErrorCellsToFallbackValues == false and "" if setErrorCellsToFallbackValues == true
For v2, I assume you would want to return "" if useNullForErrorCells == false and null if useNullForErrorCells == true
v2 specifically returns the formula in case of error [here](https://github.com/crealytics/spark-excel/blob/935a9494a2dbe15f181b6c55f0efaaf43773e2b4/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala#L86). Not sure where the code that does that for v1 is

### Steps To Reproduce

The PR will include updated tests that you can use to play with this

### Environment

```markdown
latest versions
```

cristichircu:
@nightscape @quanghgx please share your thoughts on this and the proposed fix

PR #588, opened 2022-04-29
handle errors in formula cells

cristichircu:
fix for https://github.com/crealytics/spark-excel/issues/587

For v1 the previous behavior is changed and, instead of the formula, this will return null if setErrorCellsToFallbackValues == false and "" if setErrorCellsToFallbackValues == true
For v2, if useNullForErrorCells == false this maintains the current behavior, returning the formula, and returns null if useNullForErrorCells == true
I would propose to also change [this line](https://github.com/crealytics/spark-excel/blob/935a9494a2dbe15f181b6c55f0efaaf43773e2b4/src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala#L86) to return "" in case of error for string formula to keep this consistent with v1. Not sure why anyone would want to see the formula.

### DIFF
--- a/src/test/scala/com/crealytics/spark/v2/excel/PlainNumberReadSuite.scala
+++ b/src/test/scala/com/crealytics/spark/v2/excel/PlainNumberReadSuite.scala
@@ -63,7 +63,7 @@ object PlainNumberReadSuite {
     Row("123456789012", "123456789012", "0.01"),
     Row("-0.12345678901", "0.05", "0h 14m"),
     Row(null, null, null),
-    Row("-1/0", "abc.def", null)
+    Row("#DIV/0!", "abc.def", null)
   ).asJava
 
   val expectedExcelDataNonInferSchema: util.List[Row] = List(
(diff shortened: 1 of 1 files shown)


cristichircu:
because of the change in default behavior for v1 the tests in com.crealytics.spark.excel.PlainNumberReadSuite are currently failing. I would argue that this is the correct behavior, but I get that this might be a breaking change for someone that used it like this up to now.
Any pointers about the desired behavior would be appreciated.
Thanks!

nightscape:
Hey @cristichircu, this looks reasonable to me. I don't have time to do a more thorough review, but if noone objects, I'd be happy to merge it after you have updated com.crealytics.spark.excel.PlainNumberReadSuite.
@quanghgx @christianknoepfle opinions?

cristichircu:
Had another look at how error formulas for string columns were handled in v2. In the current version the formula generating the error was returned. I updated it to return the error value, to make it consistent with errors on string columns and what users see in excel (please check the test change for an example).
@quanghgx any reason why you chose to implement the 'useNullForErrorCells' option and return the error values instead of implementing the 'setErrorCellsToFallbackValues' option and returning default values?

cristichircu:
@quanghgx @nightscape Anything I can help with to get this merged?

nightscape:
Yes, a ping 😄 
Merging now 👍

cristichircu:
:) Thank you, @nightscape !

PR #613, opened 2022-07-13
Support an option to set POI IOUtils max byte array size

pjfanning:


nightscape:
I think a new fork would make things more complicated. @pjfanning can you incorporate the changes?

src/main/scala/com/crealytics/spark/excel/WorkbookReader.scala
```diff
@@ -56,26 +56,48 @@ object WorkbookReader {
...
-    extends WorkbookReader {
+class StreamingWorkbookReader(
+  inputStreamProvider: => InputStream,
+  workbookPassword: Option[String],
+  maxRowsInMem: Int,
+  maxByteArraySize: Option[Int]
+) extends WorkbookReader {
   override protected def openWorkbook(): Workbook = {
+    maxByteArraySize.foreach { maxSize =>
+      IOUtils.setByteArrayMaxOverride(maxSize)
```
nightscape, line 99: Does this parameter work in `StreamingReader` as well?

src/main/scala/com/crealytics/spark/excel/package.scala
```diff
@@ -96,7 +97,8 @@ package object excel {
         "timestampFormat" -> timestampFormat,
         "maxRowsInMemory" -> maxRowsInMemory,
         "excerptSize" -> excerptSize,
-        "workbookPassword" -> workbookPassword
+        "workbookPassword" -> workbookPassword,
+        "maxByteArraySize" -> maxByteArraySize
```
nightscape, line 101: Nitpick: Could you move this line up above `workbookPassword`?
It is related to memory management and would fit well with `maxRowsInMemory` or `excerptSize`.

src/main/scala/com/crealytics/spark/excel/WorkbookReader.scala
pjfanning: yes - excel-streaming-reader uses some POI code under the hood

src/main/scala/com/crealytics/spark/excel/package.scala
pjfanning: done

pjfanning:
I've applied the changes requested @nigl too

RakhiSh13:
Hi is it still the same solution setting IOUtils.setByteArrayMaxOverride?

nightscape:
Nope: https://github.com/crealytics/spark-excel/pull/613/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R94

Issue #623, opened 2022-08-01
[BUG] API V2 Cannot write DataFrame with Booleans


E-HO:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hi,

Searched for "Boolenan" in the list of bugs and in the docs, but I haven't see any result to fix the issue or the question.

While trying to write some data-frame, it doesn't work anymore and "crash" when the DF to save contains columns with Boolean values, no matter if the DF is huge or very light.

Error while doing this : 

- Output Console (Notebook) : "org.apache.spark.SparkException: Job Aborted."

- Driver logs / Standard Error : 
![image](https://user-images.githubusercontent.com/16776195/182137131-e20a5af1-f741-4dc8-88c5-19b014cce511.png)

### Steps To Reproduce

df_cols = ["col_1", "max", "Boolean"] 
df_vals = [ 
  ("abc", "01", True), 
  ("def", "05", False), 
  ("ghi", "10", True), 
  ("jkl", "15", False), 
] 

df = spark.createDataFrame(df_vals, df_cols)

df.coalesce(1).write \ 
    .format("excel") \ 
    .option("header", "true") \ 
    .option("dataAddress", f"'{sheet}'!A1") \ 
    .option("usePlainNumberFormat", "true") \ 
    .mode("append") \ 
    .save("/mnt/PathToSave")

### Environment

```markdown
- Spark version: 3.2.1
- Spark-Excel version: 0.17.1
- OS: Databricks 10.4 LTS
- Cluster environment : Standard_F16S , Scala 2.12 , Python Notebook
```

pjfanning:
I made a PR (#624) and raised #625

PR #624, opened 2022-08-01
ExcelGenerator should support boolean type

pjfanning:
Relates to https://github.com/crealytics/spark-excel/issues/623

ExcelGenerator is missing cases for a large number of other data types.

### DIFF
--- a/src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
+++ b/src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
@@ -45,16 +45,17 @@ object WriteAndReadSuite {
       StructField("Extra Option 1", IntegerType, true),
       StructField("Extra Option 2", IntegerType, true),
       StructField("Extra Option 3", IntegerType, true),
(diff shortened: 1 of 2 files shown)


src/test/scala/com/crealytics/spark/v2/excel/WriteAndReadSuite.scala
```diff
@@ -72,6 +72,12 @@ object WriteAndReadSuite {
     Row(4, "2021-11-11", "2021-11-11 01:23:05"),
     Row(5, "2022-10-01", "2022-10-01 16:23:45")
   )
+
+  val expectedData_bool: List[Row] = List(Row(1, true), Row(2, false))
```
nightscape, line 76: Could you check if the tests correcly uncover the missing functionality if you add a bool to `userDefinedSchema_01`?
That would save a lot of code (the entire `write then read boolean` test).

pjfanning: @nightscape updated the test code

Issue #650, opened 2022-10-01
[BUG] Error reading XML stream when using streaming reader and V2


christianknoepfle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Loading a larger xlsx (actually the one provided in #322 ) fails with
`
com.github.pjfanning.xlsx.exceptions.ParseException: Error reading XML stream
	at com.github.pjfanning.xlsx.impl.StreamingRowIterator.getRow(StreamingRowIterator.java:126)
	at com.github.pjfanning.xlsx.impl.StreamingRowIterator.hasNext(StreamingRowIterator.java:626)
	at scala.collection.convert.Wrappers$JIteratorWrapper.hasNext(Wrappers.scala:45)
`
using 
`
        spark.read.format("excel")
        .option("path", "src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx")
        .option("header", value = false)
        .option("maxRowsInMemory", "200")
        .option("inferSchema", false)
        .load()
`

### Expected Behavior

Does not fail to load the file :)

### Steps To Reproduce

The unit test is [here:](https://github.com/christianknoepfle/spark-excel/blob/pr/closed-stream-issue/src/test/scala/com/crealytics/spark/v2/excel/RowNumberColumnSuite.scala)

The file is [here](https://github.com/christianknoepfle/spark-excel/blob/pr/closed-stream-issue/src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx)

### Environment

```markdown
- Spark version: 3.2.2
- Spark-Excel version: 0.18.0
- OS: Windows / Ubuntu (WSL2)
- Cluster environment
```

### Anything else?

First I thought that this is an issue with the streaming reader itself. I checked the file with the latest code of on github and it worked fine.

After some debugging I found the issue. 

ExcelPartitionReaderFactory.readFile calls ExcelHelper.getRows(). There the workbook gets closed too early. If I simply remove the "finally workbook.close()" the file loads fine. But the workbook stays open, so this is not the right solution. Tried scala.util.using but that didn't help

I think best would be that ExcelPartitionReaderFactory.buildReader returns a wrapped PartitionReaderWithPartitionValues  which implements the workbook.close(). 

I will try a few things and then create a PR to open a discussion on how to fix this. If you have some ideas on how to fix this, please let me know

pjfanning:
Maybe the change should be from:

```
[...]

PR #651, opened 2022-10-02
closed stream issue (issue #650)

christianknoepfle:
The issue this PR tries to fix is #650 

I introduced ExcelPartitionReaderFromIterator and moved the problematic Workbook.close() to the close() method of this class. Now spark takes care of closing the file. 

The code change is pretty straightforward. Basically I moved the code from ExcelHelper.getRows and ExcelPartitionReaderFactory.readFile to the apply() of the newly introduced class. 

In the ExcelPartitionReaderFactory.buildReader() we create an instance of the new class and pass it to PartitionReaderWithPartitionValues, which is used by spark for reading the data.
`
val fileReader = ExcelPartitionReaderFromIterator(conf, parsedOptions, file, parser, headerChecker, readDataSchema)
new PartitionReaderWithPartitionValues(fileReader, readDataSchema, partitionSchema, file.partitionValues)
`
when spark finishes reading it calls close() on PartitionReaderWithPartitionValues which in turn calls close on the fileReader. There we call close on the workbook and the issue is solved.

I am not really satisifed with the method signatures, but that was the best I could come up in the given time. If someone has an idea on how to improve it pls let me know.

The PR doesn't adress ExcelHelper.getRows(). I think this function could still cause issues, because when accessing the iterator we are reading from a closed workbook.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -80,7 +80,10 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-sql" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-hive" % testSparkVersion.value % "provided",
+  "org.apache.hadoop" % "hadoop-common" % "3.3.1" % Test,
+  "org.apache.hadoop" % "hadoop-auth" % "3.3.1" % Test,
   "org.scala-lang.modules" %% "scala-collection-compat" % "2.8.1",
+  "com.github.bigwheel" %% "util-backports" % "2.1",
   "org.typelevel" %% "cats-core" % "2.8.0" % Test,
   "org.scalatest" %% "scalatest" % "3.2.13" % Test,
   "org.scalatestplus" %% "scalacheck-1-15" % "3.2.11.0" % Test,
@@ -93,7 +96,7 @@ libraryDependencies ++= Seq(
 
 // Custom source layout for Spark Data Source API 2
 Compile / unmanagedSourceDirectories := {
-  if (testSparkVersion.value >= "3.2.2") {
+  if (testSparkVersion.value >= "3.2.0") {
     Seq(
       (Compile / sourceDirectory)(_ / "scala"),
       (Compile / sourceDirectory)(_ / "3.x/scala"),
--- a/build.sbt
+++ b/build.sbt
@@ -80,10 +80,10 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-sql" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-hive" % testSparkVersion.value % "provided",
+  // added hadoop libs to test allowing to execute the tests locally (from within intellij)
   "org.apache.hadoop" % "hadoop-common" % "3.3.1" % Test,
   "org.apache.hadoop" % "hadoop-auth" % "3.3.1" % Test,
   "org.scala-lang.modules" %% "scala-collection-compat" % "2.8.1",
-  "com.github.bigwheel" %% "util-backports" % "2.1",
   "org.typelevel" %% "cats-core" % "2.8.0" % Test,
   "org.scalatest" %% "scalatest" % "3.2.13" % Test,
   "org.scalatestplus" %% "scalacheck-1-15" % "3.2.11.0" % Test,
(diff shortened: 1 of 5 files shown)


src/test/scala/com/crealytics/spark/v2/excel/MaxNumRowsSuite.scala
```diff
@@ -0,0 +1,48 @@
...
+import com.holdenkarau.spark.testing.DataFrameSuiteBase
+import org.apache.spark.sql._
+import org.apache.spark.sql.functions.{col, max}
+import org.scalatest.wordspec.AnyWordSpec
+
+class MaxNumRowsSuite extends AnyWordSpec with DataFrameSuiteBase with LocalFileTestingUtilities {
+
+  "excel v2 and maxNumRows" can {
+
+    s"read with maxNumRows=200" in {
```
pjfanning, line 26: nit: the s interpolation is not needed

src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
```diff
@@ -135,6 +135,7 @@ class ExcelHelper private (options: ExcelOptions) {
   def getRows(conf: Configuration, uri: URI): Iterator[Vector[Cell]] = {
     val workbook = getWorkbook(conf, uri)
     val excelReader = DataLocator(options)
+    // todo this does not work with streaming reader
```
pjfanning, line 138: are in you a position to continue with the 'todo'

pjfanning:
Generally, this looks like a good way to go. If you can get your test to compile and work in all the CI builds, I think this is work merging.

pjfanning:
Generally, this looks like a good way to go. If you can get your test to compile and work in all the CI builds, I think this is work merging.

pjfanning (changes requested):

build.sbt
```diff
@@ -80,6 +80,9 @@ libraryDependencies ++= Seq(
   "org.apache.spark" %% "spark-core" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-sql" % testSparkVersion.value % "provided",
   "org.apache.spark" %% "spark-hive" % testSparkVersion.value % "provided",
+  // added hadoop libs to test allowing to execute the tests locally (from within intellij)
+  "org.apache.hadoop" % "hadoop-common" % "3.3.1" % Test,
```
pjfanning, line 84: these lib additions are causing the CI (Github Actions) build to fail

christianknoepfle: oh yeah, just found out ;) do you know why? because for local testing I had to add those

src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
christianknoepfle: not yet, I removed the TODO for now...

build.sbt
pjfanning: I'm afraid I don't know the niceties of this CI - I presume the different parallel runs are deliberately using different versions of Spark and Hadoop - so any of the jobs that need an older version of Hadoop libs are affected by this change

pjfanning:
There are also compile issues in some of the plans

```
[error] /home/runner/work/spark-excel/spark-excel/src/main/3.x/scala/com/crealytics/spark/v2/excel/ExcelPartitionReaderFromIterator.scala:17:55: type parameter InternalRow defined in class ExcelPartitionReaderFromIterator shadows class InternalRow defined in package catalyst. You may want to rename your type parameter, or possibly remove it.
[error] private[excel] class ExcelPartitionReaderFromIterator[InternalRow] private (
```

christianknoepfle:
Heck, works fine locally and fails on CI with the xml stream error message that should have been fixed now :( Will look into it later

src/test/scala/com/crealytics/spark/v2/excel/MaxNumRowsSuite.scala
christianknoepfle: removed

build.sbt
christianknoepfle: libs are only used if not running in CI. Allows to test locally without modifying the build.sbt

christianknoepfle:
@pjfanning the ci passes, but I could only acchieve this by removing the  workbook.close from ExcelHelper.getRows() for streaming workbook. So this method still gets called on V2 implementation and could cause resource issues (potentially all machines that are not my local WIndows box ;) ). I am not familiar with the code but I assume getRows is called to determine header etc. SO it needs some overhaul but I saw that you are also working on some alternative approach. Great :)

pjfanning:
> but I saw that you are also working on some alternative approach. Great :)

I'm just trying out something - feel free to review it but it's still WIP

My attempt may come to nought but I think I'm making a little progress

christianknoepfle:
I will close this PR, replaced by #653

PR #653, opened 2022-10-02
V2 streaming read (alternative approach)

pjfanning:
alternative to #651


pjfanning:
@christianknoepfle this seems to be working now - have a look if you have time

christianknoepfle:
cool :) I think I can kill my PR ;) (the core idea of the ExcelPartitionReaderFromIterator has made it in your approach but with a cleaner interface, so I am happy ;) ). 

The only thing that I didn't really liked is the fact that the CloseableIterator is not really an iterator. It is something that contains an iterator. So the name is a bit misleading (unfortunately I have no better idea for now). So finding another name or as an alternative one could implement it as an iterator
`
case class CloseableIterator[T] (val private iterator, ...)  extends Iterator[T]    ...
`
and just implement the interface forwarding the calls to the iterator. This would change all calls from it.iterator to it itself. Still you need to implement the ++ and maybe one or two other methods to make it work. 

In any case: Thanks a lot for your effort. Hope the other folks are fine with your fixes and we see a new version soon (and then hopefully we could use spark excel it in our production code)

src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
```diff
@@ -132,11 +134,24 @@ class ExcelHelper private (options: ExcelOptions) {
...
     val workbook = getWorkbook(conf, uri)
     val excelReader = DataLocator(options)
-    try { excelReader.readFrom(workbook) }
-    finally workbook.close()
+    try {
+      val rowIter = excelReader.readFrom(workbook)
+      workbook match {
+        case _: StreamingWorkbook => CloseableIterator(rowIter, Seq(workbook))
+        case _ => {
+          workbook.close()
```
christianknoepfle, line 145: why do you close the workbook here and do not hand that to CloseableIterator()? Would be more consistent to me

src/test/scala/com/crealytics/spark/v2/excel/MaxRowsReadSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+class MaxRowsReadSuite extends AnyWordSpec with DataFrameSuiteBase with Matchers {
+
+  "excel v2 and maxNumRows" can {
+
+    "read with maxNumRows=200" in {
+
+      val dfExcel = spark.read
+        .format("excel")
+        .option("path", "src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx")
+        .option("header", value = false)
```
christianknoepfle, line 32: should we test with inferSchema= true and header = true too to get greater coverage?

christianknoepfle: if you liek that, you will need the new xlsx too.. (has headers)

src/main/scala/com/crealytics/spark/v2/excel/ExcelHelper.scala
pjfanning: this workbook doesn't need to be kept open but I suppose it could be changed to work like the streaming workbook

pjfanning:
> The only thing that I didn't really liked is the fact that the CloseableIterator is not really an iterator. 

I agree that the name isn't great. I was thinking of renaming it to `SheetData` or something like and renaming the `.iterator` function to `.rowIterator`.

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -256,36 +256,48 @@ class ExcelDataSourceReader(
...
-          val r = excelHelper.getColumnNames(headerRow)
-          rows = Iterator(headerRow) ++ rows
-          r
+      try {
+        /* Prepare field names */
+        val colNames =
+          if (options.header) {
+            /* Get column name from the first row */
+            val r = excelHelper.getColumnNames(rows.rowIterator.next)
+            rows = SheetData(rows.rowIterator.drop(options.ignoreAfterHeader), rows.resourcesToClose)
```
nightscape, line 270: `SheetData` could get a `modifyIterator` method which would return a new `SheetData` with copied iterator and the same `resourcesToClose`.
Then one could do sth. like this:
```scala
rows = rows.modifyIterator(_.drop(options.ignoreAfterHeader)
```

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -256,36 +256,48 @@ class ExcelDataSourceReader(
...
+            r
+          } else {
+            /* Peek first row, then return back */
+            val headerRow = rows.rowIterator.next
+            val r = excelHelper.getColumnNames(headerRow)
+            rows = SheetData(Iterator(headerRow) ++ rows.rowIterator, rows.resourcesToClose)
```
nightscape, line 276:
```scala
rows = rows.modifyIterator(Iterator(headerRow) + _)
```

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -256,36 +256,48 @@ class ExcelDataSourceReader(
...
+            r
+          }
+
+        /* Other files also be utilized (lazily) for field types, reuse field name
+           from the first file */
+        val numberOfRowToIgnore = if (options.header) (options.ignoreAfterHeader + 1) else 0
+        rows = paths.tail.foldLeft(rows) { case (rs, path) =>
+          val newRows = excelHelper.getSheetData(conf, path)
+          SheetData(
```
nightscape, line 285: Another helpful addition would be `SheetData#append`, then the below lines could become
```scala
rs.append(newRows).modifyIterator(_.drop(numberOfRowsToIgnore))
```

src/main/2.4/scala/com/crealytics/spark/v2/excel/ExcelDataSource.scala
```diff
@@ -256,36 +256,48 @@ class ExcelDataSourceReader(
...
-      val numberOfRowToIgnore = if (options.header) (options.ignoreAfterHeader + 1) else 0
-      rows = paths.tail.foldLeft(rows) { case (rs, path) =>
-        rs ++ excelHelper.getRows(conf, path).drop(numberOfRowToIgnore)
-      }
-
-      /* Limit numer of rows to be used for schema infering */
-      rows = options.excerptSize.foldLeft(rows)(_ take _)
+        /* Limit numer of rows to be used for schema infering */
+        options.excerptSize.foreach { excerptSize =>
+          rows = SheetData(rows.rowIterator.take(excerptSize), rows.resourcesToClose)
```
nightscape, line 293:
```scala
rows = rows.modifyIterator(_.take(excerptSize))
```

src/main/3.0_3.1/scala/com/crealytics/spark/v2/excel/ExcelTable.scala
```diff
@@ -71,36 +71,49 @@ case class ExcelTable(
...
+      if (sheetData.rowIterator.isEmpty) {
+        /* If the first file is empty, not checking further */
+        StructType(Seq.empty)
+      } else {
+        /* Prepare field names */
+        val colNames =
+          if (options.header) {
+            /* Get column name from the first row */
+            val r = excelHelper.getColumnNames(sheetData.rowIterator.next)
+            sheetData = SheetData(sheetData.rowIterator.drop(options.ignoreAfterHeader), sheetData.resourcesToClose)
```
nightscape, line 86: `modifyIterator`

src/main/3.0_3.1/scala/com/crealytics/spark/v2/excel/ExcelTable.scala
```diff
@@ -71,36 +71,49 @@ case class ExcelTable(
...
+            r
+          } else {
+            /* Peek first row, then return back */
+            val headerRow = sheetData.rowIterator.next
+            val r = excelHelper.getColumnNames(headerRow)
+            sheetData = SheetData(Iterator(headerRow) ++ sheetData.rowIterator, sheetData.resourcesToClose)
```
nightscape, line 92: `modifyIterator`

src/main/3.0_3.1/scala/com/crealytics/spark/v2/excel/ExcelTable.scala
```diff
@@ -71,36 +71,49 @@ case class ExcelTable(
...
+            r
+          }
+
+        /* Other files also be utilized (lazily) for field types, reuse field name
+           from the first file */
+        val numberOfRowToIgnore = if (options.header) (options.ignoreAfterHeader + 1) else 0
+        paths.tail.foreach(path => {
+          val newRows = excelHelper.getSheetData(conf, path)
+          sheetData = SheetData(
```
nightscape, line 101: There's a lot of copied code here. I wonder at what point we DRY this and put the code that is shared for the different Spark versions into shared code.

pjfanning:
@christianknoepfle I made the changes you asked for and copied your test changes. One issue is I get dataframe sizes that are 1 less than you - for each of the 3 sub tests. I haven't investigated yet whether my code strips a row that shouldn't be stripped or if your PR overcounts.

src/main/3.2/scala/com/crealytics/spark/v2/excel/ExcelTable.scala
```diff
@@ -74,36 +74,48 @@ case class ExcelTable(
...
-        if (options.header) { /* Get column name from the first row */
-          val r = excelHelper.getColumnNames(rows.next())
-          rows = rows.drop(options.ignoreAfterHeader)
-          r
-        } else { /* Peek first row, then return back */
-          val headerRow = rows.next()
-          val r = excelHelper.getColumnNames(headerRow)
-          rows = Iterator(headerRow) ++ rows
-          r
+      try {
```
nightscape, line 82: Same here. I had not seen that there is so much overlap in the code, but we should get rid of it.

src/main/scala/com/crealytics/spark/v2/excel/SheetData.scala
```diff
@@ -0,0 +1,9 @@
+package com.crealytics.spark.v2.excel
+
+import java.io.Closeable
+
+case class SheetData[T](rowIterator: Iterator[T], resourcesToClose: Seq[Closeable] = Seq.empty) extends Closeable {
+  override def close(): Unit = {
+    resourcesToClose.foreach(_.close())
+  }
+}
```
nightscape, line 9:
```suggestion
  def modifyIterator(f: Iterator[T] => Iterator[T]): SheetData[T] = SheetData(f(rowIterator), resourcesToClose)
  def append(other: SheetData[T]): SheetData[T] = SheetData(rowIterator ++ other.rowIterator, resourcesToClose ++ other.resourcesToClose)
}
```

src/test/scala/com/crealytics/spark/excel/MaxRowsReadSuite.scala
```diff
@@ -0,0 +1,42 @@
...
+  "excel v1 and maxNumRows" can {
+
+    "read with maxNumRows=200" in {
+
+      val dfExcel = spark.read
+        .format("com.crealytics.spark.excel")
+        .option("path", "src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx")
+        .option("header", value = false)
+        // .option("dataAddress", "'Sheet1'!B7:M16")
+        .option("maxRowsInMemory", "200")
```
nightscape, line 34: As mentioned in the other PR, the integration tests actually cover streaming reads:
https://github.com/crealytics/spark-excel/blob/main/src/test/scala/com/crealytics/spark/excel/IntegrationSuite.scala#L349
If there is anything missing there, I would prefer to extend them instead of adding new tests with large binary files.

nightscape (changes requested):

nightscape:
Overall this Iterator stuff is dangerous territory. It's bitten me more than once, because at some point some code was reading from an Iterator from which another Iterator was derived...
It would be great to encapsulate that in such a way that such errors cannot occur. `SheetData` in principle could provide such isolation if the Iterator were not exposed, but it would still require some more logic to prevent double consumption.

src/main/3.2/scala/com/crealytics/spark/v2/excel/ExcelTable.scala
pjfanning: I've done some refactoring to reduce code duplication - probably more can be done

pjfanning:
> Overall this Iterator stuff is dangerous territory. It's bitten me more than once, because at some point some code was reading from an Iterator from which another Iterator was derived... It would be great to encapsulate that in such a way that such errors cannot occur. `SheetData` in principle could provide such isolation if the Iterator were not exposed, but it would still require some more logic to prevent double consumption.

For me, test coverage is the main way to manage this. The code already relies on reading off headers and things like so I'm not sure how much can be done to control that the iterators are not consumed at the wrong time.

I covered some of the topics you asked me to look at but there is more I can do.

I haven't yet looked at a way to test the changes without the new file. I'll get back to that later.

christianknoepfle:
> @christianknoepfle I made the changes you asked for and copied your test changes. One issue is I get dataframe sizes that are 1 less than you - for each of the 3 sub tests. I haven't investigated yet whether my code strips a row that shouldn't be stripped or if your PR overcounts.

you will need the new excel. I just added a header row to it. Download from [here: ](https://github.com/christianknoepfle/spark-excel/blob/pr/closed-stream-issue/src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx)

pjfanning:
> > @christianknoepfle I made the changes you asked for and copied your test changes. One issue is I get dataframe sizes that are 1 less than you - for each of the 3 sub tests. I haven't investigated yet whether my code strips a row that shouldn't be stripped or if your PR overcounts.
> 
> you will need the new excel. I just added a header row to it. Download from [here: ](https://github.com/christianknoepfle/spark-excel/blob/pr/closed-stream-issue/src/test/resources/v2readwritetest/large_excel/largefile-wide-single-sheet.xlsx)

thanks - I've updated my PR to use your latest xlsx file

src/test/scala/com/crealytics/spark/excel/MaxRowsReadSuite.scala
pjfanning: I have been unable to get any of the existing tests to reproduce the issue - the datasets are too small

nightscape:
Hey @pjfanning, I saw you picked up my changes regarding spark-testing-base to spark-fast-tests already 👍 👍 

I created a [branch which applies the integration tests to v2](https://github.com/crealytics/spark-excel/pull/657/files) as well.
It still needs some adaptations so that it doesn't assume that a single file gets written, but my hope is that it uncovers a few bugs (e.g. this one) that are still lurking in v2.

christianknoepfle:
Is there anything holding off this PR? I can say that it finally fixed one of our production issues and is key to get V2 stable. If you need help pls let me know

nightscape:
If possible, I'd like to expose the bug without adding a huge `.xlsx` file.
#657 might already expose it, but it also has/exposes other issues which might cover this one.
So I would propose the following approach:

1. @pjfanning removes the `.xlsx` file and the corresponding test from this PR.
2. I squash-merge the PR.
3. Whoever has some spare time tries to get #657 to work.
4. If that PR does not expose the bug (one probably needs to temporarily break the fix from this PR to check) we add the test with the `.xlsx`.

WDYT?

PR #664, opened 2022-10-17
Java-based Constructor for WorkbookReader

williamdphillips:
Add additional constructor for WorkbookReader which takes java hashmap as parameter for easy instantiation from python


nightscape:
Looks good to me! Thank you 👍

PR #679, opened 2022-11-07
Support Spark 3.3.X

christianknoepfle:
Spark 3.3 introduced a variety of API changes to the V2 data source read/write. Thus spark excel need to be adapted to support such API changes.

In addition spark 3.3 moved from log4j1 to log4j2. This need to be honored as well.

The folder structure for v2 is getting messy, maybe we find a way to improve that. 

@nightscape, I created the PR so you can see the current state. I will still do some more working, trying to clean up the folder structure. The unit tests all pass

Todos:
- [x] remove the test log4j properties, not used anyhow 
- [x] use spark 3.3.1 and not 3.3.0 
- [x] see if folder structure can be improved
- [x] see if the 3.3.x implementation for spark excel can be improved (code reduction)


src/main/scala/com/crealytics/spark/excel/WorkbookReader.scala
```diff
@@ -26,7 +26,8 @@ import org.apache.poi.hssf.usermodel.HSSFWorkbookFactory
 import org.apache.poi.openxml4j.util.ZipInputStreamZipEntrySource
 import org.apache.poi.util.IOUtils
 import org.apache.poi.xssf.usermodel.XSSFWorkbookFactory
-import scala.collection.JavaConverters.mapAsScalaMapConverter
+
+import scala.jdk.CollectionConverters._
```
nightscape, line 30: Just a heads-up: In my current project we just ran into a problem where including `scala-collection-compat` into the fat jar yielded errors because the included `scala` package (from `scala.jdk`) prevented the actual Scala package from being loaded. We are not bundling `scala-collection-compat` therefore I hope this won't be an error, but in case someone complains this would be the place to look at.

christianknoepfle: revert that change, did that by accident

nightscape:
@christianknoepfle I just renamed the `...v2.excel` branch to `...excel.v2`. I hope it doesn't impact your work too much.
If you want you can just ignore the change and I'll rebase your PR before merging.
The reason for the change is that with the `v2.excel` version both implementations did only have the `com.crealytics.spark` package as common parent. and this always keeps me from putting shared code into that.
So hopefully, this makes extracting code shared between v1 and v2 a little easier, at least in terms of package structure.

christianknoepfle:
> @christianknoepfle I just renamed the `...v2.excel` branch to `...excel.v2`. I hope it doesn't impact your work too much. If you want you can just ignore the change and I'll rebase your PR before merging. The reason for the change is that with the `v2.excel` version both implementations did only have the `com.crealytics.spark` package as common parent. and this always keeps me from putting shared code into that. So hopefully, this makes extracting code shared between v1 and v2 a little easier, at least in terms of package structure.

Just merged that in

christianknoepfle:
@nightscape, I am done with that PR, could pls provide feedback if ok or you need some addl changes. This week I have some spare time to work on these. BR

build.sbt
```diff
@@ -91,29 +88,49 @@ libraryDependencies ++= Seq(
...
       (Compile / sourceDirectory)(_ / "scala"),
-      (Compile / sourceDirectory)(_ / "3.x/scala"),
-      (Compile / sourceDirectory)(_ / "3.1_3.2/scala"),
-      (Compile / sourceDirectory)(_ / "3.2/scala")
+      (Compile / sourceDirectory)(_ / "3.3/scala"),
+      (Compile / sourceDirectory)(_ / "3.0_and_up/scala"),
+      (Compile / sourceDirectory)(_ / "3.1_and_up/scala"),
+      (Compile / sourceDirectory)(_ / "3.2_and_up/scala")
+    ).join.value
+  } else if (testSparkVersion.value >= "3.2.0") {
```
nightscape, line 110: Is the change from `3.2.2` to `3.2.0` intended here? Not sure why it was `3.2.2` in the first place, just making sure :)

christianknoepfle: Yes, this was intentional. It broke the pattern that we check on minor version and not on patch level

nightscape:
0.18.5 is on its way with this PR included. Thanks a lot @christianknoepfle !

Issue #696, opened 2022-12-16
[BUG] Classpath problems with ThinFatJar


zzeekk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using Spark-Excel 0.18.5 with Spark 3.3.1 in IntellJ, and get the following exception:
 **java.lang.NoSuchMethodError: org.apache.logging.log4j.util.LoaderUtil.getClassLoaders()[Ljava/lang/ClassLoader;**

The problems seems to be that Spark 3.3.1 has Log4j 2.17.2 as dependency, but Spark-Excel has shaded Log4J 2.19.0 into its jar (without renaming!). And there are many more dependencies shaded into it, even scala:
![image](https://user-images.githubusercontent.com/13715754/208137516-b7a9e382-5510-4dae-bd70-6825a7537973.png)

Shading in dependencies without renaming makes dependency management with Maven impossible. It potentially produces weird NoSuchMethodError and ClassNotFoundExceptions as conflicting versions are easily present in the classpath. Best practice is to let Maven solve dependencies, or rename shaded dependencies and remove them from the pom.

**Why is it done like that for spark-excel?** 

Fixing this should make spark-excel much more compatible in my opinion.

### Expected Behavior

1. **The default artifact published by spark-excel does not include any shaded dependency** (or they are renamed). Needed maven dependencies are listed in the pom as currently done.
2. An artifact including shaded dependencies can additionally be created and published using maven classifier. The classifier, e.g. "shaded", is then included in the published jar file, e.g. spark-excel_2.12-shaded-3.3.1_0.18.5.

### Environment

```markdown
- Spark version: 3.3.1
- Spark-Excel version: 0.18.5
- OS: Windows
- Cluster environment: IntelliJ
```

nightscape:
Hi @zzeekk, PRs to improve the packaging are very welcome!
The current build uses Mill, so the build.sc file would be the appropriate place for any fixes.

zzeekk:
Great, will look into it.

christianknoepfle:
See also #654

zzeekk:
Created PR #697, please check

zzeekk:
Hi @nightscape, i finalized my tests on release 0.18.6-beta1, looks good to me.
I have one more suggestion about the version number scheme, please see [0.18.6-beta1](https://github.com/crealytics/spark-excel/pull/697#issuecomment-1386080087)

PR #697, opened 2022-12-20
Bugfix Classpath problems (#696)

zzeekk:
Publish assembly jar (fat-jar) with classifier 'assembly'. Like that the main artifact is a thin-jar, usable as maven library dependency, having it's dependencies described in the maven pom, making proper dependency resolution work again. The fat-jar can be downloaded from maven central with classifier 'assembly' in the file name.

Move log4j2 implementation needed for running spark >=3.3.0 into scope test. Remove slf4j dependency as it is not used througout the project.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -45,7 +45,7 @@ class SparkModule(_scalaVersion: String, sparkVersion: String) extends SbtModule
     Rule.Relocate("org.apache.commons.io.**", "shadeio.commons.io.@1"),
     Rule.Relocate("org.apache.commons.compress.**", "shadeio.commons.compress.@1")
   )
-  override def extraPublish = Seq(PublishInfo(assembly(), classifier = None, ivyConfig = "compile"))
+  override def extraPublish = Seq(PublishInfo(assembly(), classifier = Some("assembly"), ivyConfig = "compile"))
 
   val sparkDeps = Agg(
     ivy"org.apache.spark::spark-core:$sparkVersion",
--- a/build.sc
+++ b/build.sc
@@ -60,7 +60,7 @@ class SparkModule(_scalaVersion: String, sparkVersion: String) extends SbtModule
 
   val poiVersion = "5.2.3"
   override def ivyDeps = {
-    val base = Agg(
+    Agg(
       ivy"org.apache.poi:poi:$poiVersion",
       ivy"org.apache.poi:poi-ooxml:$poiVersion",
       ivy"org.apache.poi:poi-ooxml-lite:$poiVersion",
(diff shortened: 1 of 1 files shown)


nightscape:
There's one issue with that: `spark-submit` and `spark-shell` cannot handle classifiers, i.e. there's no way to use the `assembly` jar there.
I think it would have to be the other way round, the assembly version being the default and the thin version having a `thin` classifier (I assume the thin version will mostly be used from SBT / Maven projects where classifiers work).
Or the two get different artifact IDs altogether. It would be good if the existing artifact name would be kept for the assembly so that users don't get something different than they expect.

zzeekk:
Hi @nightscape, publishing a fat-jar as main artifact with pom including library dependencies is against maven conventions. I strongly advise against this as it will produce classpath duplicates and problems.
On the otherside it is possible to use an assembly-jar with classifier with spark-submit by using parameter `--jars`, which can even take a URL of the corresponding artifact on maven central. Actionally assembly-jars should always be used with Spark through `--jars` option, as there is no pom / library dependencies which should be resolved.

nightscape:
Hi @zzeekk, I've changed the PR to target an identically named branch in our repo. I will merge this PR to this branch, then I can have CI build a release from it which you and other people can test.

nightscape:
Here we go: https://github.com/crealytics/spark-excel/actions/runs/3743993292
Once the build is finished, please test this in as many environments as you have available and in the best case also motivate other people to try it 😉

zzeekk:
Local unit tests with our Spark framework: OK

Databricks using maven dependency: OK
- Databricks Runtime Version: 12.0
- Cluster library config:
image
- Test result
image

Yarn: OPEN

zzeekk:
Hi @nightscape, finally i was able to tested this also on yarn / Spark 3.2.0, sorry for the delay:

**using maven dependency in a Spark applications pom.xml, creating a fat-jar and reading XML -> OK**
```
<dependency>
        <groupId>com.crealytics</groupId>
        <artifactId>spark-excel_2.12</artifactId>
        <version>3.2.3_0.18.6-beta1</version>
</dependency>
```

**using assembly jar with spark-shell -> OK**
```
wget https://repo1.maven.org/maven2/com/crealytics/spark-excel_2.12/3.2.3_0.18.6-beta1/spark-excel_2.12-3.2.3_0.18.6-beta1-assembly.jar
bin/spark-shell --jars spark-excel_2.12-3.2.3_0.18.6-beta1-assembly.jar
...
// v1 source
val df = spark.read.format("com.crealytics.spark.excel").option("header","true").load("abc.xlsx")
df.printSchema
df.show
// v2 source
val df = spark.read.format("excel").option("header","true").load("abc.xlsx")
df.printSchema
df.show
```

As you see i'm able to use Version 3.2.3_0.18.6-beta1 with Spark 3.2.0, as Spark's API is compatible and dependencies are stable within the same minor release, e.g. 3.2.x. You can also see this for other libraries, like delta lake: all delta lake 2.2.x versions are compatible with all spark versions 3.3.x, see also https://docs.delta.io/latest/releases.html.
I would therefore suggest another change and name the version 3.2_0.18.6-beta1 instead of 3.2.3_0.18.6-beta1.
WDYT?

tof38130:
Newer versions of package are still fat jars on mvn repo. Its seems that this branche has never been merge into main branch.
Are you planning to merge it in main branch ?

Thanks.

PR #720, opened 2023-03-12
Use takeWhile method from Range

EnverOsmanov:
The symptoms:
I have a file with ~1 million rows, 125 columns. It takes ~12 seconds to count lines with spark-excel's API V1 and ~2 minutes with API V2.

The issue:
1) `Range` does not contain own optimized method `filter`, that's why it uses method from `TraversableLike` which iterates over each number in range.
2) `r.getLastCellNum` evaluated for each number in range.

Here are some rough benchmarks with another file:
filter => 50 seconds
val lastCellNum => 38 seconds
withFilter => 20 seconds
takeWhile => 12 seconds
API V1 => 12 seconds
(File taken from [here](https://data.cityofchicago.org/Public-Safety/Crimes-2018/3i3m-jwuy) and manually converted to "xlsx")

PS. API V2 seems great! :)

### DIFF
--- a/src/main/scala/com/crealytics/spark/excel/v2/DataLocator.scala
+++ b/src/main/scala/com/crealytics/spark/excel/v2/DataLocator.scala
@@ -42,8 +42,9 @@ trait DataLocator {
         val r = sheet.getRow(rid)
         if (r == null) { Vector.empty }
         else {
+          val lastCellNum = r.getLastCellNum
           colInd
-            .filter(_ < r.getLastCellNum())
+            .takeWhile(_ < lastCellNum)
             .map(r.getCell(_, MissingCellPolicy.CREATE_NULL_AS_BLANK))
             .toVector
         }
@@ -52,12 +53,13 @@ trait DataLocator {
     } else {
       sheet.iterator.asScala
         .filter(r => rowInd.contains(r.getRowNum))
-        .map(r =>
+        .map { r =>
+          val lastCellNum = r.getLastCellNum
           colInd
-            .filter(_ < r.getLastCellNum())
+            .takeWhile(_ < lastCellNum)
             .map(r.getCell(_, MissingCellPolicy.CREATE_NULL_AS_BLANK))
             .toVector
-        )
+        }
     }
   }
 }


nightscape:
Hi @EnverOsmanov, thanks for the PR!
I'm slightly worried that `.takeWhile` slightly modifies the semantics to stop after having found the first non-matching index.
At the moment this doesn't matter because we're using a `Range` where the `colInd`s are sorted, but should someone change this to an unsorted `Seq[Int]` it would break silently.
Of course the real performance benefit outweighs the maybe-in-some-not-too-likely-future-scenario breakage, but if you find a fast version with identical semantics it would be nicer.
Could you maybe try the following?
```scala
val lastCellNum = r.getLastCellNum
colInd
  .iterator
  .filter(_ < lastCellNum)
```

EnverOsmanov:
If `coldInd` would be unsorted `Seq[Int]` and some columns would be missing,  it should break test cases.

Benchmarks:
.iterator.filter => 31 seconds
.view.filter => 2 minutes

[Here](https://github.com/crealytics/spark-excel/compare/fix/big-slow-files-v2...debug-enver?expand=1) is the code how I read the data.

Btw, I just checked the content of `colInd` for the file and I see that it is "1 to 16383" while `lastCellNum` is 23. The reason of big range is that I specified in "dataAddress" only the starting cell.

EnverOsmanov:
The alternative approach to avoid iteration over full `colInd` is in [API V1](https://github.com/crealytics/spark-excel/blob/eb21b38bcb11d681f2847369699ab97f7cf30c34/src/main/scala/com/crealytics/spark/excel/DataLocator.scala#L109):
```
.map(_.cellIterator().asScala.filter(c => colInd.contains(c.getColumnIndex)).toVector)
```

But I'm not exactly sure what was the idea behind the change in V2.

nightscape:
Hmm, maybe it is to be able to do the
```scala
r.getCell(_, MissingCellPolicy.CREATE_NULL_AS_BLANK)
```
@quanghgx could you chime in here?

EnverOsmanov:
If `colInd` would be unsorted `Seq[Int] we should sort it once. Otherwise we would be filtering on full collection for each row.

nightscape:
Superseded by #1032 which uses direct range clipping instead of `takeWhile`, achieving O(1) per row instead of O(actual columns). Thanks @EnverOsmanov for identifying the issue and the thorough benchmarking!

PR #726, opened 2023-03-30
Support spark session properties

williamdphillips:
Keeping PR in draft until I can resolve issues with SBT locally and am able to test this.


nightscape:
Hi @williamdphillips, I actually switched the build over to [Mill](https://com-lihaoyi.github.io/mill/mill/Intro_to_Mill.html).
The `build.sbt` is just there because I didn't dare to delete it yet.
I'll do so now 😉

williamdphillips:
Tested in Databricks using jar built from local repo and working as expected.

nightscape:
Merged, thank you for the contribution!
CI should build a snapshot jar, otherwise I can also create a release.

williamdphillips:
@nightscape please do create a release as this would be used in a production environment where SNAPSHOT versioning wouldn't be allowed

PR #728, opened 2023-04-03
Apply Spark Session Properties for Writing Files

williamdphillips:


williamdphillips:
Tested writing and is now working with changes:

image

nightscape:
Merged. Can you check if the same applies to the v2 package?

Issue #93, opened 2018-10-26
Dependency issues with Spark's built-in commons-compress


jwooden1:
I can use the library when I run spark on my local windows machine and read excel files on the same machine. However, when I upload the files to WASB on Azure and use HDInsight cluster for running spark jobs (either local or cluster mode), I get the following error: 

`java.lang.IllegalArgumentException: InputStream of class class org.apache.commons.compress.archivers.zip.ZipArchiveInputStream is not implementing InputStreamStatistics.
  at org.apache.poi.openxml4j.util.ZipArchiveThresholdInputStream.<init>(ZipArchiveThresholdInputStream.java:63)
  at org.apache.poi.openxml4j.opc.internal.ZipHelper.openZipStream(ZipHelper.java:180)
  at org.apache.poi.openxml4j.opc.ZipPackage.<init>(ZipPackage.java:104)
  at org.apache.poi.openxml4j.opc.OPCPackage.open(OPCPackage.java:298)
  at org.apache.poi.xssf.usermodel.XSSFWorkbookFactory.createWorkbook(XSSFWorkbookFactory.java:129)
  at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
  at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.lang.reflect.Method.invoke(Method.java:498)
  at org.apache.poi.ss.usermodel.WorkbookFactory.createWorkbook(WorkbookFactory.java:314)
  at org.apache.poi.ss.usermodel.WorkbookFactory.createXSSFWorkbook(WorkbookFactory.java:296)
  at org.apache.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:214)
  at org.apache.poi.ss.usermodel.WorkbookFactory.create(WorkbookFactory.java:180)
  at com.crealytics.spark.excel.ExcelRelation$$anonfun$openWorkbook$2$$anonfun$apply$4.apply(ExcelRelation.scala:66)
  at com.crealytics.spark.excel.ExcelRelation$$anonfun$openWorkbook$2$$anonfun$apply$4.apply(ExcelRelation.scala:66)
  at scala.Option.fold(Option.scala:158)
[...]

PR #741, opened 2023-05-25
Thin jar: publish assembly jar (fat-jar) with classifier 'assembly'

zzeekk:
see #696, #697

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -45,7 +45,7 @@ class SparkModule(_scalaVersion: String, sparkVersion: String) extends SbtModule
     Rule.Relocate("org.apache.commons.io.**", "shadeio.commons.io.@1"),
     Rule.Relocate("org.apache.commons.compress.**", "shadeio.commons.compress.@1")
   )
-  override def extraPublish = Seq(PublishInfo(assembly(), classifier = None, ivyConfig = "compile"))
+  override def extraPublish = Seq(PublishInfo(assembly(), classifier = Some("assembly"), ivyConfig = "compile"))
 
   val sparkDeps = Agg(
     ivy"org.apache.spark::spark-core:$sparkVersion",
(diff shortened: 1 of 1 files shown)


zzeekk:
Hi @nightscape, it would be nice if you could change to publish the default artifact as thin jar with an additional fat-jar artifact with classifier "assembly". WDYT?
Shall i try to fix the conflicts and make a PR into this branch, or do you want to have a look on it?

nightscape:
@zzeekk it would be great if you could rebase the PR and update the documentation. Please also try if the new instructions work, both in the command-line via spark-shell, as well as in a SBT/Mill/Gradle build (one should suffice).
Thank you! 👍👍

nightscape:
@zzeekk this might not be relevant for you anymore, but I've adapted the build process, so that there's an additional JAR with `-thin` classifier, and the assembly JAR remains the one without classifier. This solves the issue that in spark-shell and spark-submit - where you probably want the assembly JAR - there's no support for specifying classifiers in dependencies, whereas most build tools do.

nightscape:
@zzeekk because of this: https://github.com/nightscape/spark-excel/issues/93

nightscape:
It probably could also be solved by creating a `Map[SparkVersionMatcher, CommonCompressVersion]` (with invented types which would actually be just Strings) with the following semantic:
- If that Map contains an entry for a given `SparkVersionMatcher`, then we add this version as dependency and can thus work around #93 
- If it doesn't contain an entry we just skip the dependency altogether and rely on the version bundled with Spark.

If you want to give this a try, I'd be happy to support and review a PR.
I'm not sure though, if the tests would catch all errors, or if one would have to manually try each minor Spark version...

zzeekk:
Hi @nightscape, interestingly our project is on Spark 3.5.5, but still works with version 3.3.1_0.18.6-beta1.
So no hurry to fix something from our side, but happy to upgrade later when we moved to Spark 4.

Issue #730, opened 2023-04-14
Add support to spark 3.4


josecsotomorales:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

java.lang.AbstractMethodError: Receiver class com.crealytics.spark.excel.v2.ExcelPartitionReaderFactory does not define or inherit an implementation of the resolved method 'abstract org.apache.spark.sql.catalyst.FileSourceOptions options()' of abstract class org.apache.spark.sql.execution.datasources.v2.FilePartitionReaderFactory.
	at org.apache.spark.sql.execution.datasources.v2.FilePartitionReaderFactory.createReader(FilePartitionReaderFactory.scala:35)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.advanceToNextIter(DataSourceRDD.scala:84)
	at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.hasNext(DataSourceRDD.scala:63)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at scala.collection.Iterator$$anon$10.hasNext(Iterator.scala:460)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.hashAgg_doAggregateWithoutKey_1$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.hashAgg_doAggregateWithoutKey_0$(Unknown Source)
	at org.apache.spark.sql.catalyst.expressions.GeneratedClass$GeneratedIteratorForCodegenStage1.processNext(Unknown Source)
	at org.apache.spark.sql.execution.BufferedRowIterator.hasNext(BufferedRowIterator.java:43)
	at org.apache.spark.sql.execution.WholeStageCodegenExec$$anon$1.hasNext(WholeStageCodegenExec.scala:760)
	at org.apache.spark.sql.execution.SparkPlan.$anonfun$getByteArrayRdd$1(SparkPlan.scala:388)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2(RDD.scala:888)
	at org.apache.spark.rdd.RDD.$anonfun$mapPartitionsInternal$2$adapted(RDD.scala:888)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:52)
[...]

PR #754, opened 2023-07-23
Support Spark 3.4

christianknoepfle:
A PR for #730 

First trying to make a clean build for a targets and all tests passed. Then rework the folder structure (again)


christianknoepfle:
I moved the parameter/attribute handling of ExcelOptions into a Trait that is available to all builds. The version specific  ExcelOptions just inherit the trait and implement the version specific logic. So looks good to me. @nightscape , what do you think? 

As a side note: I think it would simplify things if we just start dropping support for older spark versions (similar to spark versioning policy). What are your thoughts on this?

nightscape:
@christianknoepfle looks good overall!
I just commented on two accidental changes. Once that's fixed, I'm happy to merge!

nightscape:
Release build for 0.19.0 is running: https://github.com/crealytics/spark-excel/actions/runs/5732141933

PR #817, opened 2023-12-11
Draft: feat: Add support for reading row is hidden

macsj200:
This change adds support to optionally read if a row is hidden or not.


src/main/scala/com/crealytics/spark/excel/v2/ExcelHelper.scala
```diff
@@ -216,6 +216,10 @@ class ExcelHelper private (options: ExcelOptions) {
       if (options.columnNameOfRowNumber.isDefined) Vector[String](options.columnNameOfRowNumber.get)
       else Vector.empty[String]
 
+    val rowIsHiddenColumn =
+      if (options.columnNameOfRowIsHidden.isDefined) Vector[String](options.columnNameOfRowIsHidden.get)
+      else Vector.empty[String]
```
nightscape, line 221:
```suggestion
    val rowIsHiddenColumn = options.columnNameOfRowIsHidden.toSeq
```

src/main/scala/com/crealytics/spark/excel/v2/ExcelInferSchema.scala
```diff
@@ -39,8 +39,17 @@ class ExcelInferSchema(val options: ExcelOptions) extends Serializable {
         Vector[StructField](StructField(options.columnNameOfRowNumber.get, IntegerType, false))
       } else Vector.empty
 
+    /* Possible StructField for row-is-hidden column */
+    val rowIsHiddenField =
+      if (options.columnNameOfRowIsHidden.isDefined) {
+        Vector[StructField](StructField(options.columnNameOfRowIsHidden.get, BooleanType, false))
+      } else Vector.empty
```
nightscape, line 46:
```suggestion
    val rowIsHiddenField = options.columnNameOfRowIsHidden.map(cn => StructField(cn, BooleanType, false)).toSeq
```

src/main/scala/com/crealytics/spark/excel/v2/ExcelInferSchema.scala
```diff
@@ -39,8 +39,17 @@ class ExcelInferSchema(val options: ExcelOptions) extends Serializable {
...
+
+    /* Header without row-is-hidden-column */
+    var dataHeader = if (options.columnNameOfRowIsHidden.isDefined) header.tail else header
+
```
nightscape, line 50:
```suggestion

    /* Header without row-is-hidden-column and row-number-column */
    val dataHeader = header.drop(options.columnNameOfRowIsHidden.size + options.columnNameOfRowNumber.size)
```

src/main/scala/com/crealytics/spark/excel/v2/ExcelHelper.scala
macsj200: I incorporated this suggestion. It caused a type error with the returned value below, but I changed it to concatenate `rowIsHiddenColumn.toVector` and it seems to compile now. I'm not sure what the difference is between Sequence and Vector, but it seems simple enough to convert from one to the other.

macsj200:
Thanks for the suggestions! I made some changes, and I'm running the test suite locally now.

nightscape:
@macsj200 there's some more room for improvement w.r.t. replacing `var`s by `val`s and `Option` handling.
Do you want to use this opportunity to flex your Scala muscle? 😃

macsj200:
Thanks again for the review. Unfortunately I don't have much scala muscle to speak of, let alone flex. I don't think this is the right time for me to work on this, perhaps I can take it up at a later point.

PR #948, opened 2025-03-31
Fix #747: Remove decimal points from integer values when usePlainNumberFormat=true.

### LINKED ISSUE #747: Numbers are parsed to numbers when enforceSchema is set to true and type being String
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

one column value can be large numbers like this 
1010675
note:
- enforceSchema is set to true; 
- the schema type was specified as String;
- usePlainNumberFormat = true
- this is similar to , but still different to an existing issue(https://github.com/crealytics/spark-excel/issues/126). 
I worry the title is different and the problem might be ignored, so raised here with explict problem title.

the current read/parse result is:
1010675.0

### Expected Behavior

1010675 string not transformed

### Steps To Reproduce

_No response_

### Environment

```markdown
- Spark version: 3.2.2
- Spark-Excel version: 3.2.2_0.18.5
- OS:
- Cluster environment
```

### Anything else?

_No response_

liewkaitsan:
## Problem Description

Fix Issue #747: Integer values should not display decimal points when usePlainNumberFormat=true.

## Changes

Modified the `PlainNumberFormat` class to handle integer values differently from decimal values:
- Enhanced the `format` method to detect integer values using BigDecimal's `stripTrailingZeros()` and `scale()` methods
- For integer values (when scale <= 0), removed the decimal point by using `toBigInteger().toString()`
- For non-integer values, preserved the original behavior using `toPlainString`
- Applied the same fix to both the core API and V2 API implementations

## Implementation Details

Instead of adding special handling in multiple places where cell values are extracted, this fix centralizes the logic in the `PlainNumberFormat` class. This ensures consistent behavior throughout the library and follows a more elegant approach as suggested by the project maintainer.

## Testing

- Existing tests continue to pass
- Added dedicated test cases for Issue #747 to verify the fix works correctly
- Confirmed both V1 and V2 APIs correctly handle integer values

## Compatibility

This change only affects the formatting of integer values and maintains backward compatibility with all existing functionality.


liewkaitsan:
> @liewkaitsan thanks for the PR 👍 Could you check if the same could be achieved by [configuring the DataFormatter](https://github.com/nightscape/spark-excel/blob/fbb852cb7dea3cb4d2429f21b574209438dd0141/src/main/scala/dev/mauch/spark/excel/DataColumn.scala#L51-L58)?
Thank you for the feedback! I've implemented your suggestion by modifying the `PlainNumberFormat` class instead of adding special handling in multiple places.

The updated implementation: [configuring the DataFormatter](https://github.com/nightscape/spark-excel/blob/fbb852cb7dea3cb4d2429f21b574209438dd0141/src/main/scala/dev/mauch/spark/excel/DataColumn.scala#L51-L58)
- Uses BigDecimal's `stripTrailingZeros()` and `scale()` to detect integer values
- Removes decimal points only for actual integer values
- Maintains the original behavior for non-integer values
- Applies the same fix to both core and V2 API implementations

Issue #944, opened 2025-03-10
[BUG] spark excel fails to read excel files with faulty dimension tag


christianknoepfle:
### Am I using the newest version of the library?

- [x] I have made sure that I'm using the latest version of the library.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

This is about excel files that contain multiple rows of data, but have a faulty dimension tag (`<dimension ref="A1"/>`). In that case `StreamingSheetReader.lastRowNum()` returns 0 despite the fact that there are more rows in the sheet.

Since spark excel iterates from first to last row, it iterates over zero rows and returns no data.

This is a follow up of the discussion I had over at [streaming-reader](https://github.com/pjfanning/excel-streaming-reader/issues/300). See there for initial suggestions and findings

### Expected Behavior

Read the complete excel file despite a faulty dimension tag.

I think the DataLocator needs to be changed so we will read all physical rows and do not rely on `lastRowNum` as mentioned by pj. Nevertheless any provided range in `dataAddress` need to be honoured. I might try to rework the code. Any suggestions welcome

### Environment

```markdown
- Spark version: 3.5.3
- Spark-Excel version: latest
- OS: Win
```

christianknoepfle:
I did some codning on it and the resolution seems simple... just as a heads-up

Issue #942, opened 2025-03-10
[BUG] Sheet Names consisting of digits are handled incorrectly


christianknoepfle:
### Am I using the newest version of the library?

- [x] I have made sure that I'm using the latest version of the library.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

We have to read an external excel file, whose sheet names are labelled "001", "002" etc.

If I provide the data address as `'001'!A1` (omiting ' ' doesn't change the behavior) it will not read data of the sheet `001` but `002`.

The reason is that `001` is interpreted as a sheet index. Since the left-most sheet `001` has index zero, we will read the sheet with index 1, which is `002`.

I think the issue is here: https://github.com/nightscape/spark-excel/blob/c4466bb5febafa8a05f0462cad2d7ab14f768213/src/main/scala/dev/mauch/spark/excel/DataLocator.scala#L96

The fix could be interchanging the first two conditions `getSheetAt()` with `getSheet()`. So first we will interpret the sheet name as is and if it fails it could be a number and we try this.

### Expected Behavior

A provided sheet name in dataAddress consisting of digits only should be interpreted as sheet name in the first place and if the sheet does not exist interpret is an index (as fallback)

### Steps To Reproduce

Adding following to TableReadSuite.scala

````scala

[excel_001.xlsx](https://github.com/user-attachments/files/19162765/excel_001.xlsx)

[excel_001.xlsx](https://github.com/user-attachments/files/19163007/excel_001.xlsx)

test("check sheet names consisting of digits only") {
    val df = readFromResources(
      spark,
      // path = "largefile.xlsx",
      path = "excel_001.xlsx",
      options = Map("dataAddress" -> "'001'!A1", "inferSchema" -> true, "header" -> true, "maxRowsInMemory" -> 1000)
    )
    df.printSchema()
    assert(df.schema.fields.length == 5) // sheet 001 has 5 columns, sheet 002 has 3 columns
    assert(
      df.first().getString(0) == "A"
    ) // sheet 001 contains "A" as first cell value, while sheet 002 contains  "WRONG"
  }
`````

### Environment

```markdown
- Spark version: 3.5.3
- Spark-Excel version: latest
- OS: Win
```

nightscape:
Sounds reasonable. Would you mind opening a PR?

Issue #938, opened 2025-02-27
Feature: Read multiple worksheets (of same structure) at once


christianknoepfle:
### Am I using the newest version of the library?

- [x] I have made sure that I'm using the latest version of the library.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

As of now I need to define the exact sheetName or provide a sheet index number to read data (via `dataAddress`).

### Expected Behavior

Idea is to provide a regex as `dataAddress`, e.g. "sheet_[0-9]+" and then all sheets that match the regex are read and returned as single Dataframe. All such sheets must match the same dataschema / structure. To mark the sheet name as regex we could surrounded the regex e.g. by "[...]" since [ is not a valid sheet name character).

As an alternative one could implement a function to query all the sheet names, find the matching ones, then call the dataframereader for each sheet individually. 

@nightscape , would that be a feature that fit spark-excel? If so, any thoughts about the implementation?

Best regards

Christian

### Environment

```markdown

```

nightscape:
Hi Christian, there were multiple requests regarding reading multiple sheets from an Excel file.
I think the Regex idea is nice. Is `/` allowed in sheet names? That is the notation for regexes in many languages.
Regarding the implementation:
In the best case we add the "round-trip", i.e. we allow writing to multiple sheets as well.
That way we can reuse the `IntegrationTest` which basically tests that you can read what you can write.

christianknoepfle:
`/` is a good idea, it is not an allowed char, so that fits.

christianknoepfle:
from implementation perspective I am wondering if this is the place to start: https://github.com/nightscape/spark-excel/blob/c272104635d4310a92ea01f80c8d72539ff94f98/src/main/scala/dev/mauch/spark/excel/v2/DataLocator.scala#L82

"Just" loop over all matching sheets and return a single iterator as beofre. The only caveat I see is that on the 2nd sheet and later we need to skip the header line. 

Your thoughts on this? Or would you see a beter approach?

nightscape:
That looks like the correct entrypoint 👍 
Before adding additional functionality, I would do a refactoring for using a `Seq[Sheet]` where a `Sheet` is currently used and just providing a single `Seq(sheet)` in the beginning. This should be doable while refactoring along green (i.e. letting the tests run frequently, maybe with `mill -w ...test`).

Then it would be a good time to think about testing:
[...]

PR #949, opened 2025-03-31
Read mulitple worksheets, bug fixes for faulty dimension tag, sheet names consisting of digits

christianknoepfle:
This PR adresses the following issues:
* #938
** Added ability to handle multiple sheets to CellRangeAddressDataLocator
** Added new option to ExcelOptionsTrait to signal that the sheet name is a regex expression
* #942
** Changed order of resolution in CellRangeAddressDataLocator.findSheet(). First try to get the sheet by name, then by index
* #944
** Fixed by removing the call to `getLastRowNum()` when using the streaming reader (see CellRangeAddressDataLocator.rowIndices() )

Tests are included

I haven't updated the CHANGELOG yet, because it seems to be auto-generated. And I do not know how to increase the version number.


christianknoepfle:
@nightscape , can you please have a look at this PR adressing the three issues?

src/main/scala/dev/mauch/spark/excel/v2/DataLocator.scala
```diff
@@ -77,23 +77,56 @@ object DataLocator {
...
+    val sheetIterators = sheets.zipWithIndex.map(it => {
+      val rowInd =
+        if (it._2 == 0 || !options.header)
+          rowIndices(it._1) // on first sheet we read header row, on all others we skip it *if* they contain header
+        else
+          rowIndices(it._1).drop(1) // skip header row for all but the first sheet
+      val colInd = columnIndices()
+      logInfo(s"Reading data from sheet ${it._1.getSheetName} with row indices $rowInd and column indices $colInd")
+      actualReadFromSheet(options, it._1, rowInd, colInd)
+    })
```
nightscape, line 102:
```suggestion
    val sheetIterators = sheets.zipWithIndex.map { case (sheet, idx) =>
      val rowInd =
        if (idx == 0 || !options.header)
          rowIndices(sheet) // on first sheet we read header row, on all others we skip it *if* they contain header
        else
          rowIndices(sheet).drop(1) // skip header row for all but the first sheet
      val colInd = columnIndices()
      logInfo(s"Reading data from sheet ${sheet.getSheetName} with row indices $rowInd and column indices $colInd")
      actualReadFromSheet(options, sheet, rowInd, colInd)
    }
```

README.md
```diff
@@ -184,8 +184,19 @@ df.write
...
+The data address consists of two portions:
+* The sheet name (optional) 
+* The cell range 
+
+For example `'My Sheet'!B3:F35` will read from the sheet `My Sheet` and the cell range `B3:F35`.  
+
+Following rules apply for the sheet name:
+* The sheet name is optional and can be omitted. In that case data is read from the first sheet (the leftmost sheet).
+* If the sheet name consists of digits only (e.g. `001`), spark excel will try to find/read from sheet named `001`. In case no sheet with this name exists, it will read the sheet with index 1 (zero-based, i.e. the second sheet from the left side).
+* If you set the spark option `sheetNameIsRegex` to `true`, the sheet name will be interpreted as a regex pattern. In this case, data of all sheets matching the regex will be read. The data schema for all such sheets must be the same.
```
nightscape, line 197: (Without yet having thought this through thoroughly) I would slightly prefer using a naming convention for the data address instead of another separate parameter.
E.g. a `/` is officially not allowed as sheet name, but regexes often written as `/sheet[0-9]/`.
So using this syntax would communicate the user intent for using a regex quite clearly.

What are your thoughts on this?

src/main/scala/dev/mauch/spark/excel/v2/DataLocator.scala
christianknoepfle: nice :)

PR #956, opened 2025-04-22
Add `assembly` classifier to uber-jar, publish normal jar

### LINKED ISSUE #852: [FEATURE] Optimize JAR size
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Hello, I was checking your library and I realised that the final jar size is quite big (30MB).
The library is very useful but it's very heavy to include it in fat jar, maybe it's possible to optimize a bit.
Checking on [maven](https://mvnrepository.com/artifact/com.crealytics/spark-excel_2.13/3.5.0_0.20.3) I noticed that:
 - you are including scala as compile dependency but I think it could be marked as provided 
 - you are including both poi-ooxml and poi-ooxml-lite (I think that only one of these should be included checking the [FAQ#3](https://poi.apache.org/help/faq.html)
 
 I don't know if these tips can help

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

bastewart:
With this change two dependency jars will be produced instead of the current one.

The current uber-jar will now have a `-assembly` suffix on the aretefact. In place of the un-classified jar will be a "thin" jar; one with just the `spark-excel` classes.

This change means end-users can avoid pulling in the uber-jar to environments where it may not be suitable, especially considering the uber-jar contains unshaded dependencies.

Resolves #852

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -71,7 +71,7 @@ trait SparkModule extends Cross.Module2[String, String] with SbtModule with Sona
     Rule.Relocate("org.apache.commons.compress.**", "shadeio.commons.compress.@1")
   )
 
-  override def extraPublish = Seq(PublishInfo(assembly(), classifier = None, ivyConfig = "compile"))
+  override def extraPublish = Seq(PublishInfo(assembly(), classifier = Some("assembly"), ivyConfig = "compile"))
 
   def publishArtifacts: T[PublishModule.PublishData] = Task {
     val publishData = super.publishArtifacts()


nightscape:
The issue is that spark-submit and spark-shell don't support specifying a classifier, which would make the plugin unusable with those two (see https://github.com/nightscape/spark-excel/issues/654#issuecomment-1305942798).
Build tools like Mill, SBT, Maven and Gradle do support classifiers, so there it would not be an issue.
Reversing the logic so that the normal JAR gets a `-thin` classifier might be an option.

PR #976, opened 2025-06-27
Added Spark 4 Support

josecsotomorales:


nightscape:
@josecsotomorales thanks for the PR!
I'm on the phone now and can't do a proper review, so here's just a high-level one: 
It looks like the 4.0_and_up directory contains a copy of the entire source code for the lower versions.
Did the API change very much between 3.x and 4.0?
Up to now, I tried to keep the version-specific code to a minimum in order to keep functionality mostly identical without having to edit duplicate code. This was annoying to set up, but allowed us to maintain compatibility with a wide range of Spark versions for quite a long time. 

Maybe that was your plan anyway, but I'll ask explicitly 🙂
Could you try to minimize the duplication of code? Any class that is an exact copy, you should be able to just throw away. 
Any class that required chances would need some consideration how to reduce the amount of duplication to a pragmatically minimum amount, e.g. by introducing a super-class with all the code that is valid for all/several versions and a subclass that is version-specific and put into the corresponding version-specific folders. 

I've also been thinking about introducing the `@ifdef` compiler plugin, which might help getting rid of the version-specific folders altogether and use conditional compilation instead. 
I didn't have time to push this through, but maybe it would be a good time now, if it helps with Spark 4 support. 

Let me know what you think 🙂

josecsotomorales:
@nightscape, sorry for the duplication, that wasn't intentional. I cleaned up the PR so only the necessary changes were added/updated. When you have time, I would appreciate your feedback. Thanks!

nightscape:
@josecsotomorales the PR looks quite good now 👍 
There seems to be an issue with encoding an `Any` type (see the CI run).
Could you have a look at that?

josecsotomorales:
@nightscape we've got another round of changes, can we run the test workflows?

nightscape:
Tests are looking good, and the diff shrank quite a bit! 👍 
Let me know once it's ready for review.

nightscape:
Great work @josecsotomorales!!
I merged it and will cut a release now 👍

nightscape:
After a few hickups, `0.31.2` is published containing support for Spark 4.
@josecsotomorales @sshpuntoff

